PR #91, opened 2023-12-04
PR - Speech to Text and Teams Extension

hunterjam:
## Purpose

* These changes include: speech-to-text via the web interface, a Teams extension, and updated README content.

## Does this introduce a breaking change?

```
[ ] Yes
[x] No
```

## Pull Request Type
What kind of change does this Pull Request introduce?

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[x] Documentation content changes
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
Follow the updated steps on the README file.
```

* Test the code

```
Follow the updated steps on the README file.
```

## What to Check
Verify that the following are valid
* When loading the web chat interface, verify the microphone icon is available. Click the icon and verify the microphone is working.

PR #92, opened 2023-12-05
Revert "Update Bicep and ARM Template file to support other Regions for Document Intelligence deployment"

gmndrg:
Reverts Azure-Samples/chat-with-your-data-solution-accelerator#76

There is a change in this PR that unfortunately breaks the deployment template. It will have to be fixed and resubmitted.


gmndrg:
@hunterjam can you please take a look so we can revert so the reported issue is fixed? This will have to be resubmitted with the fix to support any region of Form Recognizer and new log analytics workspaces.

hunterjam:
@gmndrg --- I did resolve the issue within PR #91.

gmndrg:
Thank you so much @hunterjam  I will mention PR in the issue you've raised and close this reversion PR

PR #98, opened 2023-12-08
Update BatchPushResults.py

nitin-neural:
case fix for file extension

## Purpose

* ... Files uploaded with extension not in small case is ignored by the azure function as no matching file type. No error thrown and the data is not inserted in the index.

```
[ ] Yes
[x ] No
```

## Pull Request Type

```
[x ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  deploy the code

* Test the code

```
```

## What to Check
* ... Try uploading any supported file with the filename extension in capital letters


ross-p-smith (approved):
Great - many thanks

PR #100, opened 2023-12-11
Update deployment.json

anddan1106:
fix bug when ResourcePrefix + resourceGroup().id is less than 24 chars, using take instead of substring

## Purpose

* ...

```
[ ] Yes
[ X ] No
```

## Pull Request Type

```
[ X ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

- go to [anddan1106-deploy branch]( https://github.com/anddan1106/chat-with-your-data-solution-accelerator/tree/anddan1106-deploy)
- click deploy from readme
- assign Resource Prefix to a string of 4 chars
- assign values non optional fields
- validate the deployment

* Test the code

```
```

## What to Check
* no validation error should be thrown

## Other Information
substring("string", 0, 24) require a string length of 24 or more, and will always return 24 chars. 
take("string", 24) will return at most 24 chars, otherwise original string will be returned

### DIFF
--- a/infrastructure/deployment.json
+++ b/infrastructure/deployment.json
@@ -263,7 +263,7 @@
     },
     "StorageAccountName": {
       "type": "string",
-      "defaultValue": "[substring(format('{0}str{1}', parameters('ResourcePrefix'), uniqueString(resourceGroup().id)), 0, 24)]",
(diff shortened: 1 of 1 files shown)


anddan1106:
Sorry, i was unable to recreate correctly. 
This should not be a problem since the resourceGroup().id ensures a string length > 24

PR #186, opened 2024-01-15
Authenticate using RBAC instead of API Keys for Azure Search & Azure Open AI resources

### LINKED ISSUE #41: Use managed identity, not keys
app.py currently uses keys, not managed identity. Best practice for enterprises is to use identity tokens. See my blog post here:

http://blog.pamelafox.org/2023/09/best-practices-for-openai-chat-apps-go.html

Unfortunately, the dataSources API doesn't yet support managed identity, but we're told the change is coming soon.

komalg1:
## Purpose

* Option to authenticate using RBAC roles instead of API keys for Azure Search & Azure Open AI resources.

```
[x] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

## Other Information

Partially fixes #41


code/utilities/helpers/AzureSearchHelper.py
```diff
@@ -60,14 +60,24 @@ def get_vector_store(self):
             ),
         ]
         
+        if env_helper.AZURE_AUTH_TYPE == "keys":
+            search_key = env_helper.AZURE_SEARCH_KEY
```
adamdougal, line 64: Instead of repeating the full `AzureSearch(...)` block. Can you instead just base the assignment of the `search_key` variable on the condition using [ternary operators](https://book.pythontips.com/en/latest/ternary_operators.html)? E.g:
```python
search_key = env_helper.AZURE_SEARCH_KEY if env_helper.AZURE_AUTH_TYPE == "keys else None
```
Same for the `get_conversation_logger` function below.

code/utilities/helpers/EnvHelper.py
```diff
@@ -40,13 +42,17 @@ def __init__(self, **kwargs) -> None:
         self.AZURE_OPENAI_STREAM = os.getenv('AZURE_OPENAI_STREAM', '')
         self.AZURE_OPENAI_EMBEDDING_MODEL = os.getenv('AZURE_OPENAI_EMBEDDING_MODEL', '')
          # Set env for OpenAI SDK
-        self.OPENAI_API_TYPE = "azure"
         self.OPENAI_API_BASE = f"https://{os.getenv('AZURE_OPENAI_RESOURCE')}.openai.azure.com/"
-        self.OPENAI_API_KEY = self.AZURE_OPENAI_KEY
+        if self.AZURE_AUTH_TYPE == "keys": 
+            self.OPENAI_API_KEY = self.AZURE_OPENAI_KEY
+            self.OPENAI_API_TYPE = "azure"
```
adamdougal, line 48: I think we could use a ternery operator here too.

code/utilities/helpers/EnvHelper.py
```diff
@@ -40,13 +42,17 @@ def __init__(self, **kwargs) -> None:
...
+        else:
+            self.OPENAI_API_KEY = DefaultAzureCredential(exclude_shared_token_cache_credential=True).get_token("https://cognitiveservices.azure.com/.default").token
```
adamdougal, line 50: I'm not sure having this logic in a generic `EnvHelper` is best. I think it could be confusing as sometimes the `OPENAI_API_KEY` will be set to a token instead of a key. 

Could we instead move this logic to the constructor of the `LLMHelper` which uses the key? What do you think?

adamdougal (changes requested):
A few minor comments. Other then that it looks great!

code/utilities/helpers/LLMHelper.py
```diff
@@ -10,7 +10,7 @@ def __init__(self):
         env_helper: EnvHelper = EnvHelper()
 
         # Configure OpenAI API
-        openai.api_type = "azure"
+        openai.api_type = env_helper.OPENAI_API_TYPE
         openai.api_version = env_helper.AZURE_OPENAI_API_VERSION
         openai.api_base = env_helper.OPENAI_API_BASE
         openai.api_key = env_helper.OPENAI_API_KEY
```
komalg1, line 16: @adamdougal should we add the auth_type logic here to set the api_key as key or token?

code/utilities/helpers/EnvHelper.py
```diff
@@ -40,13 +42,17 @@ def __init__(self, **kwargs) -> None:
...
+            self.OPENAI_API_TYPE = "azure_ad"
         self.OPENAI_API_VERSION = self.AZURE_OPENAI_API_VERSION
-        os.environ["OPENAI_API_TYPE"] = "azure"
+        os.environ["OPENAI_API_TYPE"] = self.OPENAI_API_TYPE
         os.environ["OPENAI_API_BASE"] = f"https://{os.getenv('AZURE_OPENAI_RESOURCE')}.openai.azure.com/"
-        os.environ["OPENAI_API_KEY"] = self.AZURE_OPENAI_KEY
+        os.environ["OPENAI_API_KEY"] = self.OPENAI_API_KEY
```
komalg1, line 55: the openai_api_key is also being set in the environ variables, so if we move the logic to LLMHelper constructor, then we need to set the environ variable os.environ["OPENAI_API_KEY"] as well which I am guessing can also become confusing because all other variables are being set here in the EnvHelper file. What do you think? @adamdougal

adamdougal: Yea that's a very good point. I'm going to backtrack and say that how you have it already is best, keeping the logic in one place!

adamdougal (approved):
This is great! Can you confirm that you've tested this is working as expected before we merge? Thanks!

lgtm

komalg1:
Yes tested it running in local

AthanasiosPavlou:
Kindly note that this is proving problematic as per above issue and kindly asking to revert (as our accelerator deployments have stopped working) :(

adamdougal:
@AthanasiosPavlou Heya, we've just merged a change to revery this issue in #212. Once the GitHub workflows have run this issue should resolve. Apologies about that

PR #187, opened 2024-01-15
updating docs for local functions debugging

komalg1:
## Purpose

* Updating the Local_Deployment.md file for debugging functions locally.

```
[ ] Yes
[x] No
```

## Pull Request Type
Updating the local deployment instructions

```
[ ] Bugfix
[ ] Feature
[x] Documentation content changes
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/docs/LOCAL_DEPLOYMENT.md
+++ b/docs/LOCAL_DEPLOYMENT.md
@@ -78,13 +78,32 @@ If you want to develop and run the batch processing functions container locally,
 
 First, install [Azure Functions Core Tools](https://learn.microsoft.com/en-us/azure/azure-functions/functions-run-local?tabs=windows%2Cportal%2Cv2%2Cbash&pivots=programming-language-python).
 
-
 ```shell
 cd code\batch
 func start
 ```
 
 Or use the [Azure Functions VS Code extension](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-azurefunctions).
+
+#### Debugging the batch processing functions locally
+Add a new file `local.settings.json` in the `batch` folder with content as below and update the `AzureWebJobsStorage` value with the storage account connection string:
+``` json
+{
+    "IsEncrypted": false,
+    "Values": {
+      "FUNCTIONS_WORKER_RUNTIME": "python",
+      "AzureWebJobsStorage": "",
+      "MyBindingConnection": "",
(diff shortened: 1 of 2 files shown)


docs/LOCAL_DEPLOYMENT.md
```diff
@@ -78,14 +78,33 @@ If you want to develop and run the batch processing functions container locally,
...
-
 ```shell
 cd code\batch
 func start
 ```
 
 Or use the [Azure Functions VS Code extension](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-azurefunctions).
 
+#### Debugging the batch processing functions locally
+Add a new file `local.settings.json` in the `batch` folder with content as below and update the `AzureWebJobsStorage` value with the storage account connection string:
```
adamdougal, line 89: Could this file be commited to the repository? A bit weird as it's a "local" file. But if it's useful for everyone it may be worth it.

docs/LOCAL_DEPLOYMENT.md
```diff
@@ -78,14 +78,33 @@ If you want to develop and run the batch processing functions container locally,
...
+      "AzureWebJobs.HttpExample.Disabled": "true"
+    },
+    "Host": {
+      "LocalHttpPort": 7071,
+      "CORS": "*",
+      "CORSCredentials": false
+    }
+  }
+```
+Execute the above shell command to run the function locally. You may need to stop the deployed function on the portal so that all requests are debugged locally.
```
adamdougal, line 106: Should we make this clearer that it's the command fromt he previous section? Maybe add a link to it in the `above shell command` text.

docs/LOCAL_DEPLOYMENT.md
komalg1: actually i can add it as local.settings.json.sample something. Ideally it should not be pushed to repo as it may contain connection string. It has been add in the .funcignore file
Also it is only helpful in case anyone wants to debug functions locally.

komalg1: added local.setting.json.sample file.

adamdougal: Ah that's a better idea!

PR #194, opened 2024-01-16
Update mimetype for streamed responses to application/json-lines

### LINKED ISSUE #42: Response types have wrong mime-type
Response currently has "event/stream" which is the mime-type for SSE, but the response is not SSE - it's a streaming HTTP response using Transfer-Encoding: chunked.

I agree with the choice of response type (as I detail in my blog post @ http://blog.pamelafox.org/2023/08/fetching-json-over-streaming-http.html ), but the mime-type should be changed. I think it can simply be dropped entirely.

Right now, the browser shows an Event Source tab, thinking that SSE events are coming in, and then none come in. If you make the change, you should be able to see the response come in correctly with no Event Source tab in Developer Tools.

cecheta:
## Purpose

Closes #42 (again)

Update mimetype to `application/json-lines`. Response now appears correctly in browser/Postman.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
* Hit `/api/conversation/azure_byod` endpoint

## What to Check
See response
Screenshot 2024-01-16 163344
Screenshot 2024-01-16 163326
Screenshot 2024-01-16 163948
Screenshot 2024-01-16 163118

### DIFF
--- a/code/app/app.py
+++ b/code/app/app.py
@@ -178,9 +178,9 @@ def conversation_with_data(request):
         return Response(json.dumps(r).replace("\n", "\\n"), status=status_code)
     else:
         if request.method == "POST":
-            return Response(stream_with_data(body, headers, endpoint))
+            return Response(stream_with_data(body, headers, endpoint), mimetype="application/json-lines")
         else:
-            return Response(None)
+            return Response(None, mimetype="application/json-lines")
 
 def stream_without_data(response):
     responseText = ""
@@ -251,9 +251,9 @@ def conversation_without_data(request):
         return jsonify(response_obj), 200
     else:
         if request.method == "POST":
-            return Response(stream_without_data(response))
+            return Response(stream_without_data(response), mimetype="application/json-lines")
         else:
-            return Response(None)
+            return Response(None, mimetype="application/json-lines")
 
 @app.route("/api/conversation/azure_byod", methods=["GET", "POST"])
 def conversation_azure_byod():


code/app/app.py
```diff
@@ -178,9 +178,9 @@ def conversation_with_data(request):
         return Response(json.dumps(r).replace("\n", "\\n"), status=status_code)
     else:
         if request.method == "POST":
-            return Response(stream_with_data(body, headers, endpoint))
+            return Response(stream_with_data(body, headers, endpoint), mimetype="application/json-lines")
         else:
-            return Response(None)
+            return Response(None, mimetype="application/json-lines")
```
pamelafox, line 183: This is for a different PR, but shouldn't this be an HTTP 405 (Method Not Allowed)? Not sure how your UI renders it.

pamelafox (approved):
Mime-type looks good!

code/app/app.py
cecheta: Yes I'm not quite sure what the original intention was here, or why the endpoint supposedly accepts GET requests, but I thought it would be best to leave it as-is and just change the mimetypes

cecheta: I've raised it in a separate issue

https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/issues/197

Issue #114, opened 2023-12-19
Index error when user queries are complex or in quick session


arun13go:
> Please provide us with the following information:
> ---------------------------------------------------------------

### This issue is for a: (mark with an `x`)
```
- [ x] bug report -> please search issues before submitting
- [ ] feature request
- [ ] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

### Minimal steps to reproduce
> After successfully uploaded the large document
> User asking queries from those docs
> Sequence of queries from the document one by one or ask ask complex queries
> UI thro UI error "An error occurred. Please try again. If the problem persist, please contact the site administrator"
> seems index out of range error in the OutputParserTool.py so look like issue in conversation chain cache

### Any log messages given by the failure
> see attached (redacted)

### Expected/desired behavior
>

### OS and Version?
> Windows 7, 8 or 10. Linux (which distribution). macOS (Yosemite? El Capitan? Sierra?)
Microsoft Windows 10 Enterprise
Version  10.0.19044 Build 19044

### Versions
>

### Mention any other details that might be useful
[redacted_error_userquery.log](https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/files/13714289/redacted_error_userquery.log)

> ---------------------------------------------------------------
> Thanks! We'll be in touch soon.

arun13go:
Hey Team, Is there update as customer is waiting for a fix. Can you please prioritise this bug if possible.

adamdougal:
I'm still working to figure out why this is happening. But this is being caused because the call to open ai here:

https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/main/code/utilities/orchestrator/OpenAIFunctions.py#L78

Is returning a stop finish reason. That causes the code to enter the else block, which does not populate the `source_documents` data:

https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/main/code/utilities/orchestrator/OpenAIFunctions.py#L104

However, in the returned message, it still includes a reference to a document e.g. `.... [doc1]`. We then look up source document 1 from the empty list and a IndexError is thrown.

adamdougal:
I think this is caused by the previous citation references e.g. `.... [doc1]` being included in the previous messages we send to the chatgpt model here:

[...]

PR #195, opened 2024-01-16
Fix IndexError when collating chat history

adamdougal:
## Purpose
This fixes a bug which causes the exception:
```
ERROR:root:Exception in /api/conversation/custom | list index out of range
Traceback (most recent call last):
  File "/workspaces/chat-with-your-data-solution-accelerator/code/app/app.py", line 283, in conversation_custom
    chat_history.append((user_assistant_messages[i]['content'],user_assistant_messages[i+1]['content']))
```

This is caused when there has been an error providing a response, and the latest message in the history is from a user, rather than the assitant. Our code assumes a user message is always followed by an assistant message.

This change removes that assumption and explitely retreives the role for each message when collating the chat history.

Required by https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/issues/114

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```


ross-p-smith:
I hate asking this - but is there a test that we can put around this?

adamdougal:
@ross-p-smith test added!

ross-p-smith (approved):
💖

Issue #215, opened 2024-01-22
Alert loop with 'copy' event listener in chat window


jtejeroc:
**/code/app/frontend/src/components/Answer/Answer.tsx**
When using chat webApp and you try to copy the conversation content, on 'copy' event, alert "*Please consider where you paste this content.*" keeps reappearing on a loop after clicking on "Accept". The "clicking" must go on incrementally with the queries sent by the user.
- One user query: alert is shown once and pops out when you click on Accept once.
- Two user queries: alert is shown, click on Accept, pops out and a new one reappears, click Accept again and pops out.
- And so on.

Additional observations:
- It does not matter how many messages are you copying.
- Loop does **NOT** restart if you clear the conversation with the broom icon.
- Loop **restarts only** if you refresh the chat webpage.

> Please provide us with the following information:
> ---------------------------------------------------------------

### This issue is for a: (mark with an `x`)
```
- [X] bug report -> please search issues before submitting
- [ ] feature request
- [ ] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

### Minimal steps to reproduce
> Send one or several messages and try to copy content.

### Any log messages given by the failure
> No.

### Expected/desired behavior
> Copy event alert to be shown only once when you copy or even just once per "chat session"; could be set to reappear if "clear chat" is clicked or webpage is refreshed.

### OS and Version?
> Windows 10 Pro (22H2).

### Versions
>

### Mention any other details that might be useful
- Copy event seems to be accumulating as the total user queries increase.
- It does not depend on how many messages do you want to copy.
- Loop does not restart even if you clear the conversation.
- Loop restarts only if you refresh the chat webpage.
> ---------------------------------------------------------------

adamdougal:
@jtejeroc Heya! This issue has now been resolved. Can you deploy the latest version and confirm it is fixed for you? Thanks

PR #230, opened 2024-01-23
Fix duplicate copy warning alert

adamdougal:
## Purpose
- It was being added to the whole page everytime an answer was returned
- This meant that if **any** content on the page was copied the alert was being shown multiple times based on how many answers there had been
- This change will mean the alert will now only be shown when an answer is copied and only once

Required by https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/issues/215

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/code/app/frontend/src/components/Answer/Answer.tsx
+++ b/code/app/frontend/src/components/Answer/Answer.tsx
@@ -14,11 +14,13 @@ import supersub from 'remark-supersub'
 interface Props {
     answer: AskResponse;
     onCitationClicked: (citedDocument: Citation) => void;
+    index: Number;
 }
 
 export const Answer = ({
     answer,
-    onCitationClicked
+    onCitationClicked,
+    index,
 }: Props) => {
     const [isRefAccordionOpen, { toggle: toggleIsRefAccordionOpen }] = useBoolean(false);
     const filePathTruncationLimit = 50;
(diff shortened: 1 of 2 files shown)


code/app/frontend/src/components/Answer/Answer.tsx
```diff
@@ -14,11 +14,13 @@ import supersub from 'remark-supersub'
 interface Props {
     answer: AskResponse;
     onCitationClicked: (citedDocument: Citation) => void;
+    index: Number;
```
cecheta, line 17:
```suggestion
    index: number;
```

code/app/frontend/src/components/Answer/Answer.tsx
```diff
@@ -57,15 +59,16 @@ export const Answer = ({
         const handleCopy = () => {
             alert("Please consider where you paste this content.");
         };
-        document.addEventListener("copy", handleCopy);
+        var messageBox = document.getElementById("message-" + index);
```
cecheta, line 62:
```suggestion
        const messageBox = document.getElementById("message-" + index);
```

cecheta, line 62: Maybe define `"message-" + index` as a variable then use it in both places

Also can use a template string

cecheta (approved):
✅

PR #247, opened 2024-01-30
Simple CI for Pull Requests

ross-p-smith:
## Purpose
Currently the repository only runs basic unit tests on a Pull Request. There is no way to know whether the typescript front end compiles or the python apps work.

This PR is an initial start to improve that process. For each PR the following steps will be carrried out.

- Pull the code
- Login to GitHub Container Registry
- Pull a cached dev container
- Build the dev container (using the cache)
- Run make commands inside the devcontainer - this means anything you can run in the devcontainer will run in the build. We currently run `make ci`
- If we are on main then push the new devcontainer image

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
```

## What to Check
When a PR is created the CI workflow runs


Makefile
```diff
@@ -0,0 +1,27 @@
...
+			include $(ENV_FILE)
+			export
+		endif
+	endif
+endif
+
+help: ## 💬 This help message :)
+	@grep -E '[a-zA-Z_-]+:.*?## .*$$' $(firstword $(MAKEFILE_LIST)) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-23s\033[0m %s\n", $$1, $$2}'
+
+ci: unittest build-frontend ## 🚀 Continuous Integration (called by Github Actions)
```
adamdougal, line 19: Should we also run formatting and linting checks as part of this?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v2 
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
```
cecheta, line 22: Does this affect you?

https://docs.opensource.microsoft.com/github/apps/permission-changes/

Makefile
ross-p-smith: This will fail the build if I add them here as another pr needs to fix the code first, but it's in a few days we will turn that on here too

.github/workflows/ci.yml
ross-p-smith: Fantastic! Yes - fixed

cecheta (approved):

PR #319, opened 2024-02-22
Upgrade langchain amd azure-search-documents

### LINKED ISSUE #316: Update Search document libraries to use 2023-10-01 Preview
Update all Search doc libraries that included 2023-07-01 Preview and update to use 2023-10-01 Preview

ross-p-smith:
## Purpose

Fixes #316

Upgrade the azure-search-documents from 11.4.0b8 to 11.4.0. I have attempted to upgrade this to 11.6 but there is compatibility issues with langchain. At the moment, there is an issue that informs me to stick to 11.4. I will investigate further. However, moving off a beta version has introduced a breaking change 

```
vector_search_configuration="default",
```
is now
```
vector_search_profile_name="myHnswProfile",
```

- I have also managed to debug all code locally and have fixed a missing environment variable
- The integration tests were also failing as a pdf that was being used has expired.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x] Other... Please describe: Upgrade Azure Search Documents
```

## Other Information
This is an attempt to eventually move to the new Azure Search API.


.github/workflows/Backend.yml
cecheta, line 1: Why are we removing these workflows?

adamdougal: We don't use the Docker images any more

cecheta: Then should we remove all of them?

adamdougal: Yea we should. Can either do it as part of this PR or a later one.

cecheta (approved):

.github/workflows/Backend.yml
ross-p-smith: Just deleted the ones that are failing the build for the moment. Think we need a strategy for build and containers which we plan for March

PR #113, opened 2023-12-18
Create OpenAI resource automatically

adamdougal:
## Purpose
- Removes a manual pre-req

## Does this introduce a breaking change?

```
[ ] Yes
[x] No
```

## Pull Request Type
What kind of change does this Pull Request introduce?

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Documentation content changes
[ ] Other... Please describe:
```

## How to Test
* Test the code
```
1. Deploy using bicep file/deployment json
```

## What to Check
Verify that the following are valid
* The app functions as expected

## Other Information
Fixes https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/issues/108

PR #410, opened 2024-03-12
Remove openai pre-req from readme

adamdougal:
- These resources are now created when running `azd up`
- This change was made in https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/113 but this section of the README was missed


ross-p-smith:
I agree that we don't need to have a model deployment, but maybe we should state that we need OpenAI enabled in the subscription - because I seem to remember you have to enable it

adamdougal:
Good point! I'll change it to state that requirement instead.

ross-p-smith (approved):

Issue #408, opened 2024-03-11
Improve unit testing code coverage.


ross-p-smith:
# Motivation 
The accelerator has a small number of Unit tests to check the validity of the application. This ticket can be closed when we have the framework in place to increase the code coverage for the solution. This ticket is not here to write tests.

# Requirements
- [ ] Add Code Coverage to CI so that we can track that every PR increases this coverage
- [ ] Fail the build (can turn off later) if the threshold decreases to encourage every PR to not make things worse

![gifs](https://media.giphy.com/media/v1.Y2lkPTc5MGI3NjExbTZnYXdrcHo0eGVkampiY2VlamdjMWVjNHphZ3Jsem56aDV6ZXdkdyZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/3orieKKmYyvUdR3RkY/giphy.gif)

PR #432, opened 2024-03-14
Add coverage to PRs

cecheta:
## Purpose

Required for #408 

This PR adds commenting to PRs with code coverage.

The commenting is done using a third-party GitHub action, [pytest-coverage-comment](https://github.com/marketplace/actions/pytest-coverage-comment). It is pinned to a specific Git SHA to prevent automatically picking up new changes to the repo which could potentially be malicious.

The commenting is handled in a separate workflow, as the token in the `Tests` workflow only has read-only permissions and cannot comment on PRs when running from forks. The separate workflow for commenting only has write permissions to pull requests, and no other permissions.

This commenting action does not have the ability to fail a run based on a threshold, therefore this will be introduced in a separate PR. The next PR will also handle commenting after test failures.

The PR also introduces caching of pip dependencies.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
```

## How to Test
It is is difficult to test as it will only run after it has been merged. I have managed to test it in my own fork.

image
image

### DIFF
--- a/.github/workflows/comment_coverage.yml
+++ b/.github/workflows/comment_coverage.yml
@@ -1,0 +1,31 @@
+name: Comment coverage
+
+on:
+  workflow_run:
+    workflows: [Tests]
+    types:
+      - completed
+
+permissions:
+  pull-requests: write
+
+jobs:
+  comment:
+    name: Comment coverage
+    runs-on: ubuntu-20.04
+    if: >
+      github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion == 'success'
+    steps:
+      - name: Download artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: coverage
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          run-id: ${{ github.event.workflow_run.id }}
+      - name: Comment coverage
+        uses: MishaKav/pytest-coverage-comment@a1fe18e2b00c64a765568e2edb9f1706eb8fc88b
+        with:
+          pytest-xml-coverage-path: coverage.xml
+          junitxml-path: coverage-junit.xml
+          issue-number: ${{ github.event.workflow_run.pull_requests[0].number }}
--- a/.github/workflows/comment_coverage.yml
+++ b/.github/workflows/comment_coverage.yml
@@ -29,3 +29,5 @@ jobs:
           pytest-xml-coverage-path: coverage.xml
           junitxml-path: coverage-junit.xml
           issue-number: ${{ github.event.workflow_run.pull_requests[0].number }}
+          coverage-path-prefix: code/
+          report-only-changed-files: true
--- a/.github/workflows/comment_coverage.yml
+++ b/.github/workflows/comment_coverage.yml
@@ -12,7 +12,7 @@ permissions:
 jobs:
   comment:
     name: Comment coverage
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     if: >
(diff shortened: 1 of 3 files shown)


ross-p-smith:
Not a reflection on the PR - which is great; more a question on the coverage number. It states 67% - that seems much higher than I was expecting. Is that 67% of code covered in the files that we actually have tests for? This will be fun as soon as someone writes a test for a file not included there, I can imagine that overall coverage number racing down not up!

ross-p-smith (approved):
I fully appreciate that things like this are hard to author until they hit main. I look forward to seeing where this goes!

.github/workflows/unittests.yml
```diff
@@ -2,24 +2,39 @@ name: Tests
...
+          architecture: x64
+          cache: pip
+          cache-dependency-path: "**/*requirements*.txt"
+      - name: Install dependencies
+        run: |
+          pip install -r code/requirements.txt -r code/dev-requirements.txt -r code/backend/requirements.txt
+      - name: Run Python tests
+        run: python -m pytest -m "not azure" --junitxml=junit-coverage.xml --cov --cov-report xml:coverage.xml
+        working-directory: code
+      - uses: actions/upload-artifact@v4
```
adamdougal, line 34: Where does this upload the coverage report too out of curiosity? How does it differentiate coverage reports from different workflow runs? Is there a retention period?

Sorry for all the questions, I've not used it before!

cecheta:
> Not a reflection on the PR - which is great; more a question on the coverage number. It states 67% - that seems much higher than I was expecting. Is that 67% of code covered in the files that we actually have tests for? This will be fun as soon as someone writes a test for a file not included there, I can imagine that overall coverage number racing down not up!

Good point, let me investigate

.github/workflows/unittests.yml
cecheta: The artifacts are tied to a specific workflow run, however if you have a second workflow using `workflow_run`, then you can still pull artifacts from the workflow that called it, which is whats's going on here.

There is a retention period for artifacts, you can either set your own value here or it will fall back to whatever the default is for the repository.

cecheta:
> Not a reflection on the PR - which is great; more a question on the coverage number. It states 67% - that seems much higher than I was expecting. Is that 67% of code covered in the files that we actually have tests for? This will be fun as soon as someone writes a test for a file not included there, I can imagine that overall coverage number racing down not up!

Much better (worse) now

image

It wasn't picking up files that weren't imported

.github/workflows/comment_coverage.yml
```diff
@@ -0,0 +1,33 @@
...
+    types:
+      - completed
+
+permissions:
+  pull-requests: write
+
+jobs:
+  comment:
+    name: Comment coverage
+    runs-on: ubuntu-20.04
```
ross-p-smith, line 15: Should this be latest? Seems an old ubuntu

.github/workflows/tests.yml
```diff
@@ -2,28 +2,42 @@ name: Tests
...
+      - opened
+      - ready_for_review
+      - reopened
+      - synchronize
   workflow_call:
 
 jobs:
   test_package:
     name: Tests
     runs-on: "ubuntu-20.04"
```
ross-p-smith, line 18: Same as above - is there a latest?

cecheta: ✅

ross-p-smith (approved):
🚀🧪

Issue #420, opened 2024-03-12
Add Stubbed Functional test suite for backend API


adamdougal:
# Motivation

To have ultimate confidence that changes are non-breaking, a stubbed functional test suite should be added. This would reduce the manual testing burden when reviewing PR's, especially dependabot. The benefits of having this in addition to the E2E test suite (https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/issues/419) are:
- Can test error scenarios
- Reduces OpenAI consumptions costs
- Can run locally

# How would you feel if this feature request was implemented?

![reacting](https://media.giphy.com/media/v1.Y2lkPTc5MGI3NjExcm50N285YjY1bmMxc3phOXNjbHdmam51NTBwYWVyMHF4bzh0dmxmMCZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/eNSqBI28KckvF3FFVM/giphy.gif)

# Requirements

- Focussed on the backend API only (suites for other components can come later)
- Include at least one happy path and one unhappy path test
- Functional tests that run against a locally started accelerator that points to mock servers for all downstream dependencies e.g. openai
- Should run for every PR

# Tasks

- [x] Create test for `/api/config` (basic test framework) - https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/438
- [x] Run tests as part of pipeline - https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/438
- [x] Allow overriding of application config in functional tests - https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/441
- [x] Decide on most appropriate mock server for OpenAI - https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/445
- [x] Introduce mock server - https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/447
- [x] Create happy path test for `/api/conversation/custom` (downstream mock framework required) - https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/504
- [x] Create unhappy path test for `/api/conversation/custom` - https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/520

cecheta:
For local storage we can look at using [Azurite](https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=visual-studio%2Cblob-storage)

[...]

PR #438, opened 2024-03-14
Add functional test suite

adamdougal:
- Add an initial test for the backend api config
endpoint to verify the test setup is working
- The setup aims to be reusable
- Add functional test task to the Makefile
- Rename the unittests.yml to tests.yml as these
now run more than just the unit tests

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   test_package:
-    name: Unit Tests
+    name: Tests
     runs-on: "ubuntu-20.04"
     steps:
         - uses: actions/checkout@v4
@@ -21,5 +21,7 @@ jobs:
         - name: Install dependencies
           run: |
             pip install -r code/requirements.txt -r code/dev-requirements.txt -r code/backend/requirements.txt
-        - name: Run Python tests
-          run: cd code/; python -m pytest -m "not azure"
+        - name: Run Unit tests
+          run: cd code/; python -m pytest -m "not azure and not functional"
+        - name: Run Functional tests
+          run: cd code/; python -m pytest -m "functional"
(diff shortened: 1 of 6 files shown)


code/tests/functional/README.md
```diff
@@ -0,0 +1,44 @@
+# Functional Tests
+
+This test suite is designed to run against local instances of the applications that constitute this project.
+
+All tests should treat the applications as close boxes, interacting only through their defined interfaces, without
```
ross-p-smith, line 5:
```suggestion
All tests should treat the applications as closed boxes, interacting only through their defined interfaces, without 
```

Makefile
```diff
@@ -16,15 +16,19 @@ endif
...
 lint: ## 🧹 Lint the code
 	@echo -e "\e[34m$@\e[0m" || true
 	@flake8 code
 
 unittest: ## 🧪 Run the unit tests
 	@echo -e "\e[34m$@\e[0m" || true
-	@cd code/ && python -m pytest -m "not azure"
+	@cd code/ && python -m pytest -m "not azure and not functional"
+
+functionaltest: ## 🧪 Run the unit tests
```
ross-p-smith, line 29:
```suggestion
functionaltest: ## 🧪 Run the functional tests
```

ross-p-smith (approved):
I love it - simple, elegant - I can see this being really useful! Will be interesting how we deal with the EnvHelper

.github/workflows/tests.yml
```diff
@@ -21,5 +21,7 @@ jobs:
         - name: Install dependencies
           run: |
             pip install -r code/requirements.txt -r code/dev-requirements.txt -r code/backend/requirements.txt
-        - name: Run Python tests
-          run: cd code/; python -m pytest -m "not azure"
+        - name: Run Unit tests
+          run: cd code/; python -m pytest -m "not azure and not functional"
```
cecheta, line 25: Could we use `working-directory` instead of `cd`?

adamdougal:
> Will be interesting how we deal with the EnvHelper

I'm going to tackle that next! 

here goes

.github/workflows/tests.yml
adamdougal: Yes good idea!

cecheta (approved):
Nice

PR #441, opened 2024-03-14
Allow the overriding of app config from functional tests

adamdougal:
- Ideally we wouldn't be setting actual environment vars
- The ideal solution would be to supply the config via a configuration file
or as startup parameters
- However, I've implemented this way for now to avoid any changes to the actual code


cecheta:
Maybe it will be clearer later, but would someone use this to override config?

cecheta (approved):

Issue #170, opened 2024-01-12
Investigate and Implement : Using Poetry to handle the requirements better


ross-p-smith:
Look at using Poetry to handle the requirements between development and runtime

-----------------------------------------------------------

### This issue is for a: (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [x] feature request
- [ ] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```
Relates to #279 

Dependabot can combine updates across multiple requirements.txt files as long as they are in the same directory, or one directory deep. Because of the current setup of requirements files in this project, there are often two separate PRs to update the same package. The CI builds fail for these PRs, as the packages cannot be installed due to a mismatch in versions.

Example : 
```python
  [2024-02-05T13:11:43.257Z] ERROR: Cannot install streamlit==1.30.0 and streamlit==1.31.0 because these package versions have conflicting dependencies.
  
  [2024-02-05T13:11:43.257Z] 
  The conflict is caused by:
      The user requested streamlit==1.31.0
      The user requested streamlit==1.30.0
  
  To fix this you could try to:
  1. loosen the range of package versions you've specified
  2. remove package versions to allow pip attempt to solve the dependency conflict
  
  
  [2024-02-05T13:11:43.257Z] ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts 
``` 

To resolve this, we either need to combine all requirements into one requirements.txt file, or change the structure of the existing requirements.txt files so that they are all in the same directory, or one directory deep.

**Minimal steps to reproduce**
Run CI job for dependabot PR

**Any log messages given by the failure**
Above

**Expected/desired behavior**
Successful CI build

PR #444, opened 2024-03-15
Migrating to Poetry from requirements.txt

komalg1:
## Purpose

Required for #170 
With this change it will be easier to maintain dependabot updates coming across requirements.txt files earlier. It combines all packages in pyproject.toml.

```
[x] Yes
[ ] No
```

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
- Get the code
- Build the dev container (all the dependencies should get installed)
- For Non dev container setup execute -> pip install .
- Run the application


.github/dependabot.yml
```diff
@@ -8,12 +8,12 @@ updates:
...
+  # Dependabot will only find pyproject.toml file in the directory
   # specified, plus one level of subdirectories, therefore two separate entries
   # are required here. When updating a dependency, ensure that each package is
   # updated to the same version across both directories.
   - package-ecosystem: "pip"
-    directory: "/code"
+    directory: "/"
     schedule:
       interval: "weekly"
     open-pull-requests-limit: 50
```
adamdougal, line 19: Is the config below on lines 20-24 still required or can it now be removed?

.github/workflows/tests.yml
```diff
@@ -24,10 +24,13 @@ jobs:
...
           cache: pip
-          cache-dependency-path: "**/*requirements*.txt"
+          cache-dependency-path: "**/pyproject.toml"
+      - name: Install Poetry
+        run: |
+          pip install poetry
       - name: Install dependencies
         run: |
-          pip install -r code/requirements.txt -r code/dev-requirements.txt -r code/backend/requirements.txt
+          pip install .
```
adamdougal, line 33: I think we also need to install the `dev-dependencies`. I _think_ this only installs the `dependencies`

pyproject.toml
```diff
@@ -0,0 +1,47 @@
...
+authors = []
+readme = "README.md"
+packages = [
+    { include = "code" }
+]
+
+[tool.poetry.dependencies]
+python = "^3.10"
+azure-functions = "1.18.0"
+streamlit = "1.30.0"
```
adamdougal, line 14: I think this should be 
```suggestion
streamlit = "1.32.0"
```
based on the current requirements.txt

pyproject.toml
```diff
@@ -0,0 +1,47 @@
...
+scipy = "1.12.0"
+transformers = "4.38.2"
+python-dotenv = "1.0.1"
+azure-ai-formrecognizer = "3.3.2"
+azure-storage-blob = "12.19.1"
+azure-identity = "1.15.0"
+flask = "3.0.2"
+openai = "1.6.1"
+langchain = "0.1.9"
```
adamdougal, line 23: I think this should be 
```suggestion
langchain = "0.1.11"
```
based on the current requirements.txt

pyproject.toml
```diff
@@ -0,0 +1,47 @@
...
+langchain-community = "0.0.24"
```
adamdougal, line 24: I think this should be 

```suggestion
langchain-community = "0.0.27"
```
based on the current requirements.txt

pyproject.toml
```diff
@@ -0,0 +1,47 @@
...
+beautifulsoup4 = "4.12.3"
+fake-useragent = "1.5.0"
+chardet = "5.2.0"
+azure-search-documents = "11.4.0"
+opencensus-ext-azure = "1.1.13"
+azure-ai-contentsafety = "1.0.0"
+python-docx = "1.1.0"
+azure-keyvault-secrets = "4.4.*"
+pandas = "2.2.1"
+pre-commit = "3.6.2"
```
adamdougal, line 37: I think this should be under `dev-dependencies`

adamdougal (changes requested):

pyproject.toml
komalg1: I picked up the packages listed under dev-requirements.txt file to be added to `dev-dependencies`. 
Should 'azure-keyvault-secrets = "4.4.*"` & `pandas = "2.2.1"` be under dev dependencies?

dist/chat_with_your_data_solution_accelerator-0.1.0-py3-none-any.whl
cecheta, line 1: These look like they should be ignored

pyproject.toml
adamdougal: Ah no just the `pre-commit` in addition to what's there currently

.vscode/tasks.json
```diff
@@ -9,13 +9,13 @@
 			"isBackground": true,
 			"dependsOn": "pip install (functions)",
 			"options": {
-				"cwd": "${workspaceFolder}/code/backend/batch"
+				"cwd": "${workspaceFolder}"
 			}
 		},
 		{
 			"label": "pip install (functions)",
```
cecheta, line 16: Can we condense these into one task, since they're all just running `pip install .`? Also, does `pip install .` still work when you're not in the root of the repo?

.vscode/tasks.json
```diff
@@ -1,33 +1,13 @@
...
-			"problemMatcher": [],
-			"options": {
-				"cwd": "${workspaceFolder}/code/backend"
-			}
-		},
 		{
 			"label": "pip install (code)",
 			"type": "shell",
-			"command": "python -m pip install -r requirements.txt",
+			"command": "pip install .",
```
adamdougal, line 7: Should this be `poetry install` instead now? Might need to rename the label. Maybe it should just be generic like `install dependencies`?

.vscode/tasks.json
```diff
@@ -1,33 +1,13 @@
 {
 	"version": "2.0.0",
 	"tasks": [
-		{
-			"type": "func",
```
adamdougal, line 5: I don't think we want to delete this one

Makefile
```diff
@@ -20,19 +20,19 @@ ci: lint unittest functionaltest build-frontend ## 🚀 Continuous Integration (
...
 
 functionaltest: ## 🧪 Run the functional tests
 	@echo -e "\e[34m$@\e[0m" || true
-	@cd code/ && python -m pytest -m "functional"
+	@ poetry run pytest -m "functional"
 
 build-frontend: ## 🏗️ Build the Frontend webapp
 	@echo -e "\e[34m$@\e[0m" || true
-	@cd code/frontend && npm install && npm run build
+	@cd frontend && npm install && npm run build
```
adamdougal, line 35: This one might still need the `cd `code/frontend` part. It's just poetry that no longer needs it.

.vscode/tasks.json
komalg1: Combined it to one and `pip install .` only works when in the root of the repo. Basically it looks for the `pyproject.toml` file.

.vscode/tasks.json
komalg1: I tested without this and it works fine. Do we need a separate task for running the function?

adamdougal: I think so for ingestion

cecheta: It looks like the task comes built-in with the VS Code extension

I have a different issue though, which is that the task isn't using the venv that poetry is installing dependencies to
image

.github/workflows/tests.yml
```diff
@@ -24,20 +24,21 @@ jobs:
...
+        run: |
+          pip install poetry
       - name: Install dependencies
         run: |
-          pip install -r code/requirements.txt -r code/dev-requirements.txt -r code/backend/requirements.txt
+          poetry install
       - name: Run Unit tests
-        working-directory: ./code
-        run: python -m pytest -m "not azure and not functional" --junitxml=coverage-junit.xml --cov=. --cov-report xml:coverage.xml
+        run: poetry run pytest -m "not azure and not functional" --junitxml=coverage-junit.xml --cov=. --cov-report xml:coverage.xml
```
cecheta, line 35: We'll need to move the `.coveragerc` file into the root of the repo, and update the path inside the file for the tests directory

.vscode/tasks.json
ross-p-smith: Note that any name changes to your tasks need to be reflected in the launch.json that call these tasks. e.g. `launch Admin site` has a `preLaunchTask` that it cannot find now

.vscode/tasks.json
ross-p-smith: Today I learnt - I've just got Functions attached and it even asked me what the Storage account was! Looks like they have improved the extension a lot

dist/chat_with_your_data_solution_accelerator-0.1.0-py3-none-any.whl
ross-p-smith, line 1: These need removing from the commit

dist/chat_with_your_data_solution_accelerator-0.1.0.tar.gz
ross-p-smith, line 1: Needs removing

adamdougal (approved):
lgtm

.vscode/tasks.json
```diff
@@ -7,31 +7,22 @@
 			"command": "host start",
 			"problemMatcher": "$func-python-watch",
 			"isBackground": true,
-			"dependsOn": "pip install (functions)",
+			"dependsOn": "poetry install",
 			"options": {
-				"cwd": "${workspaceFolder}/code/backend/batch"
+				"cwd": "${workspaceFolder}"
```
cecheta, line 12:
```suggestion
				"cwd": "${workspaceFolder}/code/backend/batch"
```

If you add this back, and then within VS Code you change the Python  interpreter to the one from Poetry, then it works 😄

superhindupur (approved):

.vscode/tasks.json
komalg1: Will check this and raise a new PR. Merging this PR for now.

PR #446, opened 2024-03-15
Fail build if coverage is too low

### LINKED ISSUE #408: Improve unit testing code coverage.

cecheta:
## Purpose
Closes #408 

This PR fails the build if the coverage is lower than the current coverage in the main branch. It does this by downloading the coverage artifact from the latest build of `main`, and extracting the coverage from the XML report. It then feeds this into the pytest command with the `--cov-fail-under` flag. If the coverage cannot be obtained from the main branch, it should default to 0.

The coverage from main is rounded down to the nearest integer, as the coverage obtained from main is usually to 2dp (e.g. 33.95%), while the coverage from pytest is to a much higher precision (33.948339483%), meaning that even if there were no changes to Python code, 33.948339483% < 33.95% would fail the build.

Also, this PR allows the coverage to be commented on the PR even if tests fail, and fixes bug when running from forks: https://github.com/orgs/community/discussions/25220.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
```

## How to Test
I have tested this in my fork of this repo.

image

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -18,6 +18,7 @@
 		"vscode": {
 			"extensions": [
 				"github.vscode-pull-request-github",
+				"github.vscode-github-actions",
 				"ms-azuretools.azure-dev",
 				"ms-azuretools.vscode-azurefunctions",
 				"ms-azuretools.vscode-bicep",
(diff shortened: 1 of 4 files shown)


cecheta:
Note: Commenting isn't working from a fork, I'll look into it

PR #447, opened 2024-03-15
Introduce functional test mock server

adamdougal:
- Based on the ADR https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/main/docs/design/adrs/2024-03-15-use-pytest-httpserver-as-a-mock-server-for-functional-testing.md
- Added `OPENAI_API_BASE` to config endpoint for ease of testing
- Allowed `OPENAI_API_BASE` to be set via an env var
  - Previously this was only generated based on resource name but we
  need more control to set the mock server url
- Add unit tests to `EnvHelper` for this new functionality

Requried by https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/issues/420

### DIFF
--- a/code/tests/functional/backend_api/conftest.py
+++ b/code/tests/functional/backend_api/conftest.py
@@ -2,6 +2,7 @@ from multiprocessing import Process
 import socket
 import time
 import pytest
+from pytest_httpserver import HTTPServer
 import requests
 from app import app
 from tests.functional.backend_api.app_config import AppConfig
@@ -19,8 +20,20 @@ def app_url(app_port: int) -> int:
 
 
 @pytest.fixture(scope="module")
-def app_config() -> AppConfig:
-    return AppConfig()
+def mock_httpserver(make_httpserver):
+    """
+    This is required as the default pytest httpserver fixture is scoped at the function level
+    """
+    print("Starting HTTP Mock Server")
+    server = make_httpserver
+    yield server
+    print("Stopping HTTP Mock Server")
+    server.clear()
+
+
+@pytest.fixture(scope="module")
+def app_config(mock_httpserver: HTTPServer) -> AppConfig:
+    return AppConfig({"OPENAI_API_BASE": mock_httpserver.url_for("/")})
 
 
 @pytest.fixture(scope="module", autouse=True)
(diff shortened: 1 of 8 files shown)


code/tests/test_app.py
```diff
@@ -11,7 +11,11 @@ def test_returns_correct_config(self):
         response = app.test_client().get("/api/config")
 
         assert response.status_code == 200
-        assert response.json == {"azureSpeechKey": None, "azureSpeechRegion": None}
+        assert response.json == {
+            "azureSpeechKey": None,
+            "azureSpeechRegion": None,
+            "OPENAI_API_BASE": "https://.openai.azure.com/",
```
ross-p-smith, line 17: It's probably correct - but what's the significance of the . at the start of the URI?

adamdougal: It's because there's no `AZURE_OPENAI_RESOURCE` env var set for this test.

ross-p-smith (approved):
Ran all the tests locally - love the introduction of monkeypatch!

code/tests/test_app.py
cecheta: I feel like it would be good to have some sort of config/yaml file that sets a baseline of env vars for all the tests, then each individual test could remove/reconfigure the env vars as appropriate, but that could probably be done later

code/tests/functional/backend_api/conftest.py
```diff
@@ -19,8 +20,20 @@ def app_url(app_port: int) -> int:
...
-    return AppConfig()
+def mock_httpserver(make_httpserver):
+    """
+    This is required as the default pytest httpserver fixture is scoped at the function level
+    """
+    print("Starting HTTP Mock Server")
+    server = make_httpserver
+    yield server
+    print("Stopping HTTP Mock Server")
+    server.clear()
```
superhindupur, line 31: Could you please explain what this setup method is doing? I'm a bit confused by the start, then stop all within the same method.

superhindupur (approved):

code/tests/functional/backend_api/conftest.py
adamdougal: Heya! So this is how pytest recommeneds you create setup/teardown functions for it's tests https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended.

It runs everything before the `yield` as setup, and everything after as teardown. Adding server after `yield server` tells it what to return as part of the fixture (if anything).

Issue #417, opened 2024-03-12
Sprint Champion


ross-p-smith:
# Motivation
Dependabot is scheduled to run weekly for GitHub Actions, Python and TypeScript modules. This results in a large number of Pull Requests created that are often duplicated; this will be fixed by #279. We also do not have high test coverage on this repo. We will need someone to perform some housekeeping on this repo each week.

# Requirements
- This issue will be assigned to a new "Sprint Champion" each Sprint
- Pick a module to be updated by dependabot, write some unit/integration (delete as appropriate) tests that test that package. 
- If there are already tests for that module - then great.
- Update all of the references of the module
- All housekeeping (chores) can be linked to this issue.
- The Sprint champion will also be responsible for categorising new issues created from members of the OSS world and making sure that there is enough detail in them to be taken into the "Ready" state.

PR #450, opened 2024-03-18
Upgrade langchain

ross-p-smith:
## Purpose
Both `langchain` and `langchain-community` needed upgrading. `langchain-community` needs to use `langchain-openai` and then this causes us to upgrade the version of `openai`.

The new versions of these packages no longer use the environment variable OPENAI_API_BASE, but rather insist on using AZURE_OPENAI_ENDPOINT

I have re-ordered the environment variables in the bicep files so that they are a bit easier to see and find.

Relates to #417 

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x] Other... Please describe:
```


infra/main.bicep
ross-p-smith, line 1: The changes here are just order

code/backend/batch/utilities/helpers/EnvHelper.py
```diff
@@ -83,7 +83,8 @@ def __init__(self, **kwargs) -> None:
         # Otherwise, keys are obtained from environment variables.
         if self.AZURE_AUTH_TYPE == "rbac":
             self.AZURE_SEARCH_KEY = None
-            self.AZURE_OPENAI_KEY = ""
+            self.AZURE_OPENAI_KEY = ""  # TODO: See if this is still used.
```
adamdougal, line 86: Left over TODO

code/backend/batch/utilities/helpers/EnvHelper.py
```diff
@@ -96,22 +97,28 @@ def __init__(self, **kwargs) -> None:
...
+            f"https://{self.AZURE_OPENAI_RESOURCE}.openai.azure.com/",
         )
+        os.environ["AZURE_OPENAI_ENDPOINT"] = self.AZURE_OPENAI_ENDPOINT
+        os.environ["AZURE_OPENAI_API_KEY"] = self.AZURE_OPENAI_API_KEY
         self.OPENAI_API_TYPE = "azure" if self.AZURE_AUTH_TYPE == "keys" else "azure_ad"
         self.OPENAI_API_KEY = self.AZURE_OPENAI_KEY
         self.OPENAI_API_VERSION = self.AZURE_OPENAI_API_VERSION
         os.environ["OPENAI_API_TYPE"] = self.OPENAI_API_TYPE
-        os.environ["OPENAI_API_BASE"] = self.OPENAI_API_BASE
         os.environ["OPENAI_API_KEY"] = self.OPENAI_API_KEY
```
adamdougal, line 122: Is this still needed now we're setting `AZURE_OPENAI_API_KEY`?

code/tests/utilities/test_LangChainAgent.py
```diff
@@ -0,0 +1,63 @@
+from unittest.mock import MagicMock
+
+from backend.batch.utilities.orchestrator.LangChainAgent import LangChainAgent
+from backend.batch.utilities.common.Answer import Answer
+
+
+class Testing_LangChainAgent(LangChainAgent):
+    def __init__(self) -> None:
```
adamdougal, line 8: Nice way of overriding values when you can't provide them via the constructor! Do we need to call `super().__init__()` here though? I can see we're doing it in a few places, including the `LangChainAgent`,

code/backend/batch/utilities/helpers/EnvHelper.py
ross-p-smith: I was scared to remove it! Will do some testing

code/tests/utilities/test_LangChainAgent.py
ross-p-smith: If I call super.__init then I will also create a whole set of connected server objects which makes it hard to test. The idea was to not initialise the base class and just to test these two methods.

.gitignore
```diff
@@ -413,5 +413,6 @@ temp/
 .whl
 .tar.gz
```
cecheta, line 414:
```suggestion
*.whl
*.tar.gz
```

I know it's unrelated, but could you update this?

cecheta (approved):

Issue #419, opened 2024-03-12
Add E2E test suite


adamdougal:
# Motivation

To have ultimate confidence that changes are non-breaking, an E2E test suite should be added. This would reduce the manual testing burden when reviewing PR's, expecially dependabot.

# How would you feel if this feature request was implemented?

![nice](https://media.giphy.com/media/v1.Y2lkPTc5MGI3NjExNHhkem9haGx6NjV5MGF0ODNpNHVwcGgzZHVpZjFhajVrYzR2MThhZiZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/ZYKLgA1pB1WSPQASCs/giphy.gif)

# Requirements

- E2E tests that run against a deployed fully integrated version of the accelerator
- Only add a test for the default setup, where the front end calls `/api/conversation/custom`
  - This may require us to also test the uploading of documents via the admin application 
- Should run for every PR but only after the repository maintainers have approved the run due to security concerns https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#controlling-changes-from-forks-to-workflows-in-public-repositories

# Tasks

_To be filled in by the engineer picking up the issue_

- [x] Create ADR for which framework to use, and pick one. 
- [x] Create scaffolding for first E2E test.
- [x] Write the first E2E test where the front end calls `/api/conversation/custom`
- [ ] Set up the test to run on merge and on PR. - blocked on #563. This work is covered in #564.

PR #464, opened 2024-03-18
Create ADR for using cypress as a framework for e2e testing.

superhindupur:
## Purpose
Adds an ADR for the selection of framework for UI e2e testing - Cypress. Required by #419.

```
[ ] Yes
[x ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x ] Documentation content changes
[ ] Other... Please describe:
```

## How to Test
N/A, just verify whether the ADR makes sense based on the pros/cons.


docs/design/adrs/template.md
```diff
@@ -26,16 +26,6 @@
...
-### Positive Consequences
-
-* [e.g., improvement of quality attribute satisfaction, follow-up decisions required, …]
-* …
-
-### Negative Consequence
-
-* [e.g., compromising quality attribute, follow-up decisions required, …]
-* …
-
```
superhindupur, line 38: Removed these sections as they're covered by the pros/cons of the chosen option.

ross-p-smith:
My initial thought was not to test the UI, but rather have something like pytest that was firing a rest call to ingest a file and then calling the various endpoints to check that files had hit storage, then check the index and then ask a question all over rest. Maybe we should align.

adamdougal (approved):
Sounds good to me. From a quick google there's a Cypress GitHub action too, so we may be able to integrate this into the pipeline pretty nicely.

ross-p-smith (dismissed):

PR #479, opened 2024-03-19
Upgrade keyvault secrets library

ross-p-smith:
## Purpose
Upgrade the KeyVault library. Cleanest way to test this was to ensure that we wrote a helper method that encapsulated the SecretClient so that it could be tested.

Also found some areas that should have been using the new AZURE_OPENAI_ENDPOINT env that langchain demands.

Renamed the `Testing_LangChainAgent` class so that pytest doesn't try and collect it

Relates to #417 

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x] Other... Please describe: chore
```


code/backend/batch/utilities/helpers/EnvHelper.py
```diff
@@ -189,3 +169,46 @@ def check_env():
         for attr, value in EnvHelper().__dict__.items():
             if value == "":
                 logging.warning(f"{attr} is not set in the environment variables.")
+
+
+class SecretHelper:
```
cecheta, line 174: 😍

code/tests/utilities/test_SecretHelper.py
```diff
@@ -0,0 +1,46 @@
...
+    # then
+    assert actual_value == expected_value
+
+
+def test_get_secret_returns_value_from_secret_client_when_use_key_vault_is_true():
+    # given
+    secret_name = "MY_SECRET"
+    expected_value = "my_secret_value"
+    secret_helper = SecretHelper()
+    secret_helper.USE_KEY_VAULT = True
```
cecheta, line 25: Instead of setting this variable directly, could we monkeypatch the USE_KEY_VAULT env var? Would that work? Seems closer to how it would work in practice

ross-p-smith: Added comment in code

PR #504, opened 2024-03-20
Add functional test for /conversation/custom endpoint

adamdougal:
- Allow the disabling of app insights and loading
config from blob storage to make testing easier
- Switch to `Thread` from `Process` for running the app
  - This is to stop pytest calling the session scoped fixure multiple times
  - It also fixes an issue where we were not seeing app logs
- Setup the mock server as https due to the azure libraries enforcing it
  - This introduced additional fixtures and the `trustme` dependency
- Commit the encoding file for tiktoken
  - This is to stop it making a http call over the internet to retrive it

Still to come:
- Downstream failure tests
- Downstream request body verification


code/backend/batch/utilities/helpers/ConfigHelper.py
```diff
@@ -70,13 +70,19 @@ def __init__(self, logging: dict):
...
-            blob_client = AzureBlobStorageClient(container_name=CONFIG_CONTAINER_NAME)
-            config = blob_client.download_file("active.json")
-            config = Config(json.loads(config))
-        except Exception:
-            print("Returning default config")
-            config = ConfigHelper.get_default_config()
+        env_helper = EnvHelper()
+        config = ConfigHelper.get_default_config()
+
+        if env_helper.LOAD_CONFIG_FROM_BLOB_STORAGE:
```
ross-p-smith, line 76: lgtm - this has been a problematic area to test for lots of things - so thanks!

code/backend/batch/utilities/helpers/EnvHelper.py
```diff
@@ -128,6 +128,10 @@ def __init__(self, **kwargs) -> None:
             "AZURE_FORM_RECOGNIZER_KEY"
         )
         # Azure App Insights
+        self.APPINSIGHTS_ENABLED = (
+            os.getenv("APPINSIGHTS_ENABLED", "True").lower() == "true"
+        )
+
         self.APPINSIGHTS_CONNECTION_STRING = os.getenv(
```
ross-p-smith, line 135: There seems to be more than 1 AppInsights connection string if you have seen - do you think you can tidy that whilst here? If it is too big then feel free to create a new issue

code/backend/batch/utilities/helpers/ConfigHelper.py
adamdougal: It caused the tests to be suuuuuper slow

code/backend/batch/utilities/helpers/EnvHelper.py
adamdougal: I can't seem to find it. I only see `APPINSIGHTS_CONNECTION_STRING`

code/backend/batch/utilities/helpers/EnvHelper.py
```diff
@@ -128,6 +128,10 @@ def __init__(self, **kwargs) -> None:
             "AZURE_FORM_RECOGNIZER_KEY"
         )
         # Azure App Insights
+        self.APPINSIGHTS_ENABLED = (
```
tanya-borisova, line 131: It could be nice to have this logic defined a single time, rather than manually process each value. I did [something similar](https://github.com/microsoft/rag-experiment-accelerator/blob/development/rag_experiment_accelerator/config/environment.py#L77) in rag-experiment-accelerator recently, if this seems helpful I could create a task for this refactoring

code/backend/batch/utilities/helpers/ConfigHelper.py
```diff
@@ -70,13 +70,19 @@ def __init__(self, logging: dict):
...
+            try:
+                blob_client = AzureBlobStorageClient(
+                    container_name=CONFIG_CONTAINER_NAME
+                )
+                config_file = blob_client.download_file("active.json")
+                config = Config(json.loads(config_file))
+            except Exception:
```
tanya-borisova, line 83: I don't have much context here but do we want to be any more specific about the exception we're catching here?

code/tests/functional/backend_api/resources/README.md
```diff
@@ -0,0 +1,9 @@
+# Why is this here?
+
+The file [9b5ad71b2ce5302211f9c61530b329a4922fc6a4](9b5ad71b2ce5302211f9c61530b329a4922fc6a4) is required to stop the
```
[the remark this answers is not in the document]
adamdougal: It depends entirely on the [tiktoken ](https://pypi.org/project/tiktoken/) library which is used by openai. They have an issue open for it [here](https://github.com/openai/tiktoken/issues/232) so if that gets resolved hopefully this will go away entirely.

code/tests/utilities/test_EnvHelper.py
```diff
@@ -26,3 +26,105 @@ def test_openai_base_url_uses_env_var_if_set(monkeypatch: MonkeyPatch):
 
     # then
     assert actual_openai_endpoint == expected_openai_endpoint
+
```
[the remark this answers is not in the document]
adamdougal: Yea unfortunately it's following the convention of the codebase. It'd be nice to maybe go through and make it all PEP8 standard and consistent in a separate PR.

code/tests/utilities/test_EnvHelper.py
```diff
@@ -26,3 +26,105 @@ def test_openai_base_url_uses_env_var_if_set(monkeypatch: MonkeyPatch):
...
+
+def test_load_config_from_blob_storage_true_when_unset():
+    # given
+    expected_load_config_from_blob_storage = True
+
+    # when
+    actual_load_config_from_blob_storage = EnvHelper().LOAD_CONFIG_FROM_BLOB_STORAGE
```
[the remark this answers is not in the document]
adamdougal: Quite possibly! Let me give that a try 👍

code/tests/functional/backend_api/conftest.py
```diff
@@ -1,69 +1,171 @@
...
+    httpserver.check()
 
 
-def stop_app(proc: Process):
-    print("Shutting down application")
-    proc.terminate()
-    proc.join()  # Wait until the process is fully shut down
-    print("Application shut down")
+def start_app(app_port: int) -> Thread:
+    print(f"Starting application on port {app_port}")
```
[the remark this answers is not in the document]
adamdougal: Good point! I've not used logging with pytest before, but let me give it a go. Shouldn't require much setup if any 👍

code/backend/batch/utilities/helpers/EnvHelper.py
adamdougal: Good point! Let me make that change.

code/backend/batch/utilities/helpers/ConfigHelper.py
adamdougal: I think it wants to catch any and all Exceptions and fallback if they occur. This change is just rearranging what's there currently though.

code/tests/functional/backend_api/resources/README.md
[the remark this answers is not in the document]
tanya-borisova: Got it thanks, maybe we should add the link to the issue to this README file so that the next person can check if it can be removed

code/backend/batch/utilities/helpers/EnvHelper.py
tanya-borisova: Didn't want to make your PR bigger than it needs to be, happy to create an issue for this instead if helpful

adamdougal: I've made a small change to have a helper for env var booleans. But an issue for the rest would be good!

code/tests/functional/backend_api/resources/README.md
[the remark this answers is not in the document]
adamdougal: Good point, I've added that now!

code/tests/utilities/test_EnvHelper.py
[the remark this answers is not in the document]
tanya-borisova: https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/issues/514

tanya-borisova (approved):

code/backend/batch/utilities/helpers/EnvHelper.py
```diff
@@ -164,6 +170,9 @@ def is_chat_model(self):
             return True
         return False
 
+    def get_env_var_bool(self, var_name: str, default: str = "True") -> bool:
```
ross-p-smith, line 173: Is there a simple unit test that we can add around this - (glad you pulled this into a function btw!) :-)

ross-p-smith (approved):

code/backend/batch/utilities/helpers/EnvHelper.py
tanya-borisova: https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/issues/515

code/backend/batch/utilities/helpers/EnvHelper.py
adamdougal: I think this is covered by the existing unit tests.

Issue #378, opened 2024-03-04
Provide better Guidance on how to get the project working from clean install


ross-p-smith:
The guidance on getting the solution working from a clean install is causing confusion and needs to be tidied up.

Fixes: #373 

### This issue is for a: (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [ ] feature request
- [x] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

### Minimal steps to reproduce
Follow the instructions from the main readme.md

Lawndemon:
I second this - I need to be able to use already provisioned resources given the limited regional availability of OpenAI. I'm working on modifying the main.bicepparam file but having an variables file that could be used in place of the environment variables configured during deploy would make this much easier to work with.

ross-p-smith:
@Lawndemon - can I just understand what your requirements are here so I can see if we can address them?
- Do you want to deploy some resources in different regions than others? Can you give an example?
- You want to reuse existing resources? You can do something like `azd env set MY_ENV_KEY value` and that will write it to the default `.env` file. We could then read it in the bicepparams. Is that what you mean?

PR #505, opened 2024-03-20
AZD can use containers instead of code for deployment

### LINKED ISSUE #339: Dual deployment strategy. (azd and containers)
# Motivation

"Chat with your data" Solution accelerator currently lacks a "one-click deployment" option in the main branch, which is a critical feature for streamlined setup and user experience. 

convo

# Requirements

The one-click deployment is a significant ease-of-use feature that can help accelerate adoption and improve the initial experience for new users of the "Chat with your data" Solution accelerator. By simplifying the deployment process, we can lower the barriers to entry and help users focus on leveraging the solution's capabilities rather than being bogged down by setup procedures.

# Tasks

- [ ] Reintroduce the build and push of the 3 containers on a build of main
- [ ] Configure the App Service host bicep files to be configurable whether they deploy code or container.

- Ensuring that the ARM templates and deployment scripts in the ross/acr branch are up-to-date and compatible with the main branch.
- Updating the main branch documentation to include the one-click deployment instructions.
- Testing the merged feature thoroughly to ensure it works as expected in the main branch environment.

ross-p-smith:
## Purpose
This work enables an environment variable `AZURE_APP_SERVICE_HOSTING_MODEL` to be set which can be `code` or `container`. By default it will be container, which will pull the 3 packages Admin, Frontend, or Backend from a public Container Registry. This will greatly speed up the deployment as the code will already be built and App Service will just need to pull the container.

This has been tested by deploying a fresh instance, ingesting a document and then asking a question.

image

Fixes #339 

This fix will introduce new App Service names, so I recommend people run `azd down` and then `azd up` in order to get the cleanest build.

I will also add the "Deploy to Azure" back in as part of #378 when this hits main

```
[x] Yes
[] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```


infra/main.bicep
```diff
@@ -158,6 +158,12 @@ param principalId string = ''
 ])
 param authType string
 
+@allowed([
+  'code'
+  'container'
+])
+param hostingModel string
```
cecheta, line 165: Could this default to `container`?

infra/main.bicepparam
```diff
@@ -10,3 +10,5 @@ param principalId = readEnvironmentVariable('AZURE_PRINCIPAL_ID', 'principal_id'
 param useKeyVault = bool(readEnvironmentVariable('USE_KEY_VAULT', 'true'))
 
 param authType = readEnvironmentVariable('AZURE_AUTH_TYPE', 'keys')
+
+param hostingModel = readEnvironmentVariable('AZURE_APP_SERVICE_HOSTING_MODEL', 'container')
```
cecheta, line 14: Is this value what is used when someone deploys from the command line? If so, should it instead default to `code`?

.github/workflows/build-docker-admin.yml
```diff
@@ -0,0 +1,28 @@
+name: Admin Web App Docker Image
```
cecheta, line 1: Maybe we should only run these after `CI` finishes successfully?

Makefile
```diff
@@ -44,3 +44,9 @@ deploy: azd-login ## 🚀 Deploy everything to Azure
 	@azd deploy web --no-prompt
 	@azd deploy function --no-prompt
 	@azd deploy adminweb --no-prompt
+
+docker-build: ## 🐳 Build the Docker image
```
cecheta, line 48: What's the need for this? We're not calling it in the workflow, but the fact that it's hardcoded to `fruoccopublic.azurecr.io` seems like we wouldn't want to run it locally either?

infra/main.bicep
```diff
@@ -11,7 +11,7 @@ param resourceToken string = toLower(uniqueString(subscription().id, environment
 param location string
 
 @description('Name of App Service plan')
-param hostingPlanName string = '${environmentName}-hosting-plan-${resourceToken}'
+param hostingPlanName string = 'hosting-plan-${resourceToken}'
```
cecheta, line 14: How does this renaming affect resources that are already deployed?

superhindupur:
Reviewed this with Ross over a teams call. Here is a summary of what we discussed:

- The three dockerfiles are added for admin, frontend and backend services, since before this change, the deployment was code-based only.
- The main.bicep file and related bicep files contain changes that introduce the parameter needed for deciding between code/container deployments. There is also a bugfix in there that shortens the names of the generated app services (unrelated to this change)
- role.bicep was removed since it contained duplicate code/config. 

Testing:

- Ross tested this by running the github actions from his branch and the app was deployed successfully using the containers in the registry. (Containers were pushed to the registry from his branch too, which was essentially the same as main except the infra configuration).

Based on the above, I am happy to approve this PR.

superhindupur (dismissed):

Makefile
ross-p-smith: I've removed these as was just helpful whilst debugging - but we don't call them and you are right nobody knows the key to push them to that registry

infra/main.bicep
ross-p-smith: I believe that bicep does not manage state of changed resources, so I think that we are going to get newly create resources in peoples resource groups. I deially we need a changelog to say that breaking chnages are coming - what do you think?

infra/main.bicep
ross-p-smith: I'm following the pattern that is used for `authType` above. The bicepparam file will pass in a default of container.

infra/main.bicepparam
ross-p-smith: I've added an extra step in the README. I want the people who dont read the docs to ultimately click "Deploy to Azure" use containers and those that tinker with code to explicitly change it. In the future we could actually supply a Container Registry as part of the build - but trying to keep the scope down.

infra/main.bicep
ross-p-smith: @cecheta I've pushed and only changed the app services

Makefile
```diff
@@ -45,8 +45,6 @@ deploy: azd-login ## 🚀 Deploy everything to Azure
 	@azd deploy function --no-prompt
 	@azd deploy adminweb --no-prompt
 
-docker-build: ## 🐳 Build the Docker image
```
superhindupur: Don't we need this anymore?

ross-p-smith: Github Actions carries out the build rather than locally - this was me testing locally. We don't actually "need" it so removed it

superhindupur (approved):

infra/main.bicep
cecheta: It looks like a number of them are still changing?

image

cecheta (approved):
Nice one :)

PR #519, opened 2024-03-22
Verify the downstream calls being made by /conversations/custom

adamdougal:
- To ensure we are making the correct calls, with the correct api keys, api version, index information and request bodies
- Implemented using a generic request matcher for now
  - This should be added to the mock server library by https://github.com/csernazs/pytest-httpserver/issues/293
- Added extra tests rather then combining al the verifications in one test
  - This is so that the failure reason should be clearer
  - I was concerned that this would significantly increase the test run time however it made no difference and some runs after were actually quicker then if it was a single test
 
### DIFF
--- a/code/tests/functional/backend_api/request_matching.py
+++ b/code/tests/functional/backend_api/request_matching.py
@@ -1,4 +1,3 @@
-import logging
 from pytest_httpserver import HTTPServer
 from werkzeug.wrappers import Request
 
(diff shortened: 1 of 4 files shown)


code/tests/functional/backend_api/request_matching.py
```diff
@@ -0,0 +1,81 @@
+from pytest_httpserver import HTTPServer
+from werkzeug.wrappers import Request
+
+
+class RequestMatcher:
```
tanya-borisova, line 5: Did you say you opened an issue for pytest_httpserver? Might be useful to link it here, so that the next person knows when this code can be deprecated

tanya-borisova (approved):

cecheta (approved):

ross-p-smith (approved):

PR #537, opened 2024-03-25
Create env before deploying

cecheta:
## Purpose

* Create env before deploying, so env var can be set. Previously, the env was being created implicitly when provisioning

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
* Test after merge

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,8 @@ azd-login: ## 🔑 Login to Azure with azd and a SPN
 
 deploy: azd-login ## 🚀 Deploy everything to Azure
 	@echo -e "\e[34m$@\e[0m" || true
-	@azd env set AZURE_APP_SERVICE_HOSTING_MODEL code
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -40,7 +40,8 @@ azd-login: ## 🔑 Login to Azure with azd and a SPN
 
 deploy: azd-login ## 🚀 Deploy everything to Azure
 	@echo -e "\e[34m$@\e[0m" || true
-	@azd env set AZURE_APP_SERVICE_HOSTING_MODEL code
+	@azd env new ${AZURE_ENV_NAME}
```
ross-p-smith, line 43: AZURE_ENV_NAME is a GitHub setting - now we are not persisting the environment - maybe we could use a commit number so we dont end up being so dependent on the current value. (Not for this PR!)

PR #539, opened 2024-03-25
Removed environmentName from the start of resources

### LINKED ISSUE #427: Bug - Deployment fails if user has uppercase chars in the RG name, README does not give guidance
# Describe the bug
Following the deployment instructions from Terminal on my desktop within VS Code ... the deploy ran for approx 20+ minutes only to end in a failure. The reported message is that a child component could not be created with uppercase characters. I had furnished a resource group name as "CWYD-BobTestQA-20240313" and apparently that is a problem. After adjusting to lowercase I was next greeted with a deploy fail (after 20+ mins) stating that one of the components provisioned failed due to string length. I believe this may have been due to the "rg-" prefix and "random string guid" type suffix that is appended. Reached 65 chars in total where the cap is 64 chars max.

# Expected behavior
README should pre-warn users about uppercase characters and string length if they will result in a deployment failure. Even better is to adjust the code to force to lowercase so the deployment can proceed.

# How does this bug make you feel?
:(

---

# Debugging information

## Screenshots
image

## Logs

If applicable, add logs to help the engineer debug the problem.

---

# Tasks

- [ ] Task 1
- [ ] Task 2
- [ ] ...

ross-p-smith:
## Purpose
Fixes #427 

When we went to azd the names of the resources changed to be in the format: `${environmentName}-resource-${resourcePrefix}`. The environment name is taken from the user and therefore can cause problems with some older Azure Resources as it allows special characters and upper/lower case characters. Whereas the ResourcePrefix is generated from the subscription, environmentName and location to be unique and lowercase.

This change is BREAKING for people with existing resources as azd will not rename resources, but create new ones with the new name, because the resource group still uses the environment name (correctly imo). 

We could implement this and respect existing resources, but I would prefer to just fix this and keep the code base smaller. Discuss!

```
[x] Yes
[ ] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
Deploy it in your favourite way


cecheta:
I think I'm still in the same place as yesterday, we know that there are a number of people already using this solution and this change will create a number of duplicate resources. I think we should keep these changes behind a flag which defaults to the current naming scheme for now, and maybe put a big deprecation warning on the README that the default will be changed/the flag will be removed some time in the future.

ross-p-smith:
I can understand where you are coming from. Proposal:
1) New deployments from ARM (Deploy from Azure) button are implicitly "new" deployments and should use "USE_NEW_NAMING=true".
2) People who are deploying from `azd up` from an existing code base will not want to deploy newly named resources, and will not have that environment variable (so false)
3) People who are creating a new `azd env` will also want "USE_NEW_NAMING=true"
4) We can put a disclaimer in the readme that shows if you want to maintain the old name that you need to set "USE_NEW_NAMING=false"

Thoughts @cecheta

superhindupur:
The discussion about the duplicate resources is an interesting one. I see where you're coming from @cecheta but we might be overthinking this. What is the downside to creating duplicate resources, could you confirm? (So I know I am not missing anything)

I am not a huge fan of the USE_NEW_NAMING as it will create debt to clean up. We need to justify this if possible. Thanks!

ross-p-smith:
One problem that can occur is `azd` actually uses a tag on an App Service to know where to deploy the code to. We "could" change this name as part of this PR so that even if we do get duplicates, we only deploy code to the new one.

ross-p-smith:
I've kept this focussed around the Search Service which is where this bug originated from. Please look in the `main.biceparam` file. I've added some logic to read the environment values which are only there if someone has deployed or is using an existing environment. If I run `azd provision` with an existing `AZURE_ENV_NAME` of "ui" then the result is

```text
Resources:

  Skip     : Resource group          : rg-ui
  Create   : Cognitive Service       : contentsafety-32btpgndh24ag
  Create   : Cognitive Service       : formrecog-32btpgndh24ag
  Create   : Cognitive Service       : speech-32btpgndh24ag
  Skip     : Cognitive Service       : ui-contentsafety-32btpgndh24ag
  Skip     : Cognitive Service       : ui-formrecog-32btpgndh24ag
  Skip     : Cognitive Service       : ui-openai-32btpgndh24ag
  Skip     : Cognitive Service       : ui-speechservice-32btpgndh24ag
  Create   : Application Insights    : appinsights-32btpgndh24ag
  Skip     : Application Insights    : ui-appinsights-32btpgndh24ag
  Modify   : Key Vault               : kv-32btpgndh24ag
  Modify   : Log Analytics workspace : la-32btpgndh24ag
  Create   : Portal dashboard        : dash-appinsights-32btpgndh24ag
  Skip     : Portal dashboard        : dash-ui-appinsights-32btpgndh24ag
  Skip     : Search service          : ui-search-32btpgndh24ag
  Skip     : Storage account         : str32btpgndh24ag
  Create   : App Service plan        : hosting-plan-32btpgndh24ag
  Skip     : Web App                 : backend-32btpgndh24ag
  Modify   : Web App                 : web-32btpgndh24ag
  Skip     : Web App                 : web-32btpgndh24ag-admin
```

Notice it skips updating the Search Service. It reads the .env file and uses what is set.

Now if you run this on a new env (`azd env new 427`), it creates brand new services
```text
  Resources:

  Create : Resource group          : rg-427
  Create : Cognitive Service       : contentsafety-pp2ma7nswvzg4
  Create : Cognitive Service       : formrecog-pp2ma7nswvzg4
  Create : Cognitive Service       : openai-pp2ma7nswvzg4
  Create : Cognitive Service       : speech-pp2ma7nswvzg4
  Create : Application Insights    : appinsights-pp2ma7nswvzg4
  Create : Key Vault               : kv-pp2ma7nswvzg4
  Create : Log Analytics workspace : la-pp2ma7nswvzg4
  Create : Portal dashboard        : dash-appinsights-pp2ma7nswvzg4
  Create : Search service          : search-pp2ma7nswvzg4
  Create : Storage account         : strpp2ma7nswvzg4
  Create : App Service plan        : hosting-plan-pp2ma7nswvzg4
  Create : Web App                 : web-pp2ma7nswvzg4
```

infra/main.bicepparam
```diff
@@ -11,4 +11,20 @@ param useKeyVault = bool(readEnvironmentVariable('USE_KEY_VAULT', 'true'))
...
+
+// The following are being renamed to align with the new naming convention
+// we manipulate existing resources here to maintain backwards compatibility
+
+// We need the resourceToken to be unique for each deployment (copied from the main.bicep)
+var subscriptionId = readEnvironmentVariable('AZURE_SUBSCRIPTION_ID', 'subscription_id')
+param resourceToken = toLower(uniqueString(subscriptionId, environmentName, location))
+
+// Retrieve the Search Name from the Search Endpoint (if it exists)
+var azureAISearcEndpoint = readEnvironmentVariable('AZURE_SEARCH_SERVICE', 'https://./')
```
cecheta, line 24:
```suggestion
var azureAISearchEndpoint = readEnvironmentVariable('AZURE_SEARCH_SERVICE', 'https://./')
```

infra/main.bicepparam
```diff
@@ -11,4 +11,20 @@ param useKeyVault = bool(readEnvironmentVariable('USE_KEY_VAULT', 'true'))
...
+var searchSerivceName = split(substring(azureAISearcEndpoint, 8, length(azureAISearcEndpoint) - 9), '.')[0]
```
cecheta, line 25: Is `length(azureAISearcEndpoint) - 9` the number of characters we want to extract? If so, I think we can leave that value off and it will default to the end of the string

It's a bit hard to understand what it's doing

cecheta (approved):
🎉

PR #557, opened 2024-03-27
Don't upload node_modules to App Service

### LINKED ISSUE #540: Don't upload node_modules to App Service
# Motivation

We use `npm` to build the static code for the frontend, but once the static files are built, we no longer need `node_modules`. By uploading it to App Service, we are making the package much larger than it needs to be.

🤏

# Requirements

- Remove `node_modules` from deployment package

# Tasks

ross-p-smith:
## Purpose
Fixes #540

When we package the `web` service we zip the whole `code` folder which includes tests and node modules.

Below are the differences in zip files
```text
-rw-r--r-- 1 vscode vscode 76694284 Mar 27 21:17 /tmp/chat-with-your-data-solution-accelerator-web-azddeploy-1711574224.zip
-rw-r--r-- 1 vscode vscode  2313542 Mar 27 21:29 /tmp/chat-with-your-data-solution-accelerator-web-azddeploy-1711574955.zip
```

What we do now is 
1) Create a new `dist` folder - which is git ignored
2) Amend the npm build to output to this folder
3) Amend poetry to dump the requirements to this folder
4) Copy just the code that we want to this folder
5) For good measure don't copy the tests when we use docker too!

## How to Test
Tested manually
image


azure.yaml
```diff
@@ -9,16 +9,17 @@ services:
     project: ./code
     language: py
     host: appservice
+    dist: ./dist
     hooks:
       prepackage:
         windows:
           shell: pwsh
-          run: poetry export -o requirements.txt;cd ./frontend;npm install;npm run build;
+          run: cd frontend;npm install;npm run build;cd ..;poetry export -o dist\requirements.txt;cp app.py dist;cp -r backend dist
```
cecheta, line 17: Just checking this works in PowerShell?

azure.yaml
```diff
@@ -9,16 +9,17 @@ services:
...
           interactive: true
           continueOnError: false
         posix:
           shell: sh
-          run: poetry export -o requirements.txt;cd ./frontend;npm install;npm run build;
+          run: cd frontend;npm install;npm run build;cd ..;poetry export -o dist/requirements.txt;cp app.py dist;cp -r backend dist
```
cecheta, line 22: These are getting a bit long, thoughts on putting them in their own scripts?

azure.yaml
ross-p-smith: Yes it does!

azure.yaml
ross-p-smith: Now in a script

cecheta (approved):
Nice :)

PR #577, opened 2024-04-02
Unit tests for app.py

### LINKED ISSUE #562: Add unit tests to `/api/conversation/azure_byod` endpoint

cecheta:
Closes #562 

## Purpose

* Adds unit tests to `app.py`, increasing coverage
* `/api/conversation/azure_byod` now returns JSON response when using data + not streaming

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x] Other... Please describe: Tests
```

## How to Test
* `poetry run pytest -m "not azure and not functional"`

### DIFF
--- a/code/tests/test_app.py
+++ b/code/tests/test_app.py
@@ -1,6 +1,6 @@
 import json
 import os
-
+import pytest
 from unittest.mock import MagicMock, Mock, patch
 
 from app import app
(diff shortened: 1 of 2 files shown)


code/tests/test_app.py
```diff
@@ -258,16 +261,32 @@ def setup_method(self):
             ],
         }
```
superhindupur, line 263: Is there a way to extract the env_helper_mock setup to be common for all test cases, and have each test case only override the values if needed? I'm asking because we're needing to write so much env_helper setup for each test case.

code/tests/test_app.py
```diff
@@ -283,12 +302,120 @@ def test_converstation_azure_byod_returns_correct_response_when_streaming_with_d
...
         )
 
+        request_body = mock_session.post.call_args[1]["json"]
+        request_headers = mock_session.post.call_args[1]["headers"]
+
+        assert request_body["data_sources"][0]["parameters"]["authentication"] == {
+            "type": "api_key",
+            "key": "some-search-key",
+        }
+        assert request_headers["api-key"] == "some-api-key"
```
adamdougal, line 318: Could you use the already defined variables in the assertions? E.g.
```suggestion
        assert request_headers["api-key"] == self.openai_api_key
```
Should make the tests easier to maintain in the future.

cecheta: I've updated the majority of them, hopefully that's fine

code/tests/test_app.py
cecheta: I've put it into a fixture, what do you think?

superhindupur: Amazing, thanks for looking into this!

superhindupur (approved):

PR #582, opened 2024-04-03
Refactor app.py

cecheta:
Required for #553 

## Purpose
It is becoming increasingly difficult to test `app.py` due to the side effects that occur when `app.py` is imported, such as `EnvHelper()`. This PR moves the creation of the app into a `create_app()` function in its own `create_app.py` file, meaning that the module can be imported multiple times without any side effects. Tests are updated to accommodate the new change.

Note that a file named `app.py` with a top-level variable `app` is required for the application to run in App Service.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
* Run unit/functional tests
* Run app locally
* Run app in Azure
* Build Dockerfile and run locally


code/app.py
```diff
@@ -301,82 +282,104 @@
...
+            }
 
-        response_obj = {
-            "id": "response.id",
-            "model": env_helper.AZURE_OPENAI_MODEL,
-            "created": "response.created",
-            "object": "response.object",
-            "choices": [{"messages": messages}],
-        }
+            return jsonify(response_obj), 200
```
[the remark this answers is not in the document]
cecheta: This already exists on `main`, it is showing up here due to the refactor

code/app.py
```diff
@@ -1,381 +1,7 @@
...
-        return (
-            jsonify(
-                {
-                    "error": "Exception in /api/conversation/custom. See log for more details."
-                }
-            ),
-            500,
-        )
-
+app = create_app()
```
superhindupur, line 4: I'm not sure how this fixes the EnvHelper() side effect. This is moving the EnvHelper() call into create_app(), which is invoked here anyway. 

What am I missing?

adamdougal: I believe the difference is that we're now importing a function in tests:
`from create_app import create_app`
Instead of a variable:
`from app import app`

Previously, just importing the app would invoke the `EnvHelper` to load all vars. Now I believe it will only invoke the `EnvHelper` when the `create_app` function is called, giving us much more control. Do correct me if I've misunderstood though @cecheta!

superhindupur: Awesome, thanks for the explanation!

superhindupur (approved):

code/app.py
cecheta: Exactly, thanks @adamdougal 😄

Issue #317, opened 2024-02-21
Enable CWYD to accept user input as speech in multiple languages. 


ross-p-smith:
# Motivation

To allow users to chat with their data using speech in addition to just text. This speech input should support the following non-English languages: Swiss German: gsw-CH, German: de-CH, French: fr-CH, Italian: it-CH. 

# How would you feel if this feature request was implemented?
This will empower users to not only use speech as an input, but they can do it in multiple languages. 

# Requirements

- Users should be able to use speech as an input in non English languages to chat with their data.
- CWYD should respond in the same language as the input when the input is non-English.
- The non-English languages to be supported are:  Swiss German: gsw-CH, German: de-CH, French: fr-CH, Italian: it-CH

# Tasks

- [x] #549
- [x] Add frontend unit tests to the tests github action. #703
- [x] Accept the languages as a config input during deployment, rather than hardcoding them. #660
- [x] Edit the CWYD prompt to respond to the user in the same language as the input. (this is already in the prompt, and will be reintroduced via #645 )

PR #586, opened 2024-04-04
Add multi-language support for speech input on web.

superhindupur:
## Purpose
Required by #549. Without these changes, the chat frontend only recognizes english.  This PR adds support for multi-lingual speech recognition from the Chat frontend. Currently the additional languages are hardcoded into the frontend code, however, the plan under #317 is to move this list of languages to the deployment configuration in the near future. 

This PR also introduces the first frontend unit test with the required scaffolding and support to extend this test suite. Hereafter the expectation for frontend changes will be that they are accompanied by one or more unit tests. 

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
Deploy CWYD, and speak through the microphoine input in one of the following languages: French, German, Italian. Your words should be converted correctly to text. 

To run the frontend unit tests, checkout the code and run 
```
make unittest-frontend
```


code/frontend/.mocharc.json
```diff
@@ -0,0 +1,9 @@
+{
```
superhindupur, line 1: https://typestrong.org/ts-node/docs/recipes/mocha/ is the official recipe for typescript + mocha. 

We need ts-node for mocha to work with typescript. We are setting the loader to ESM so that it works with the ECMAScript modules in our codebase.

code/frontend/src/pages/chat/Chat.tsx
```diff
@@ -168,12 +167,12 @@ const Chat = () => {
...
-          serviceRegion
-        );
-        const audioConfig = AudioConfig.fromDefaultMicrophoneInput();
-        const recognizer = new SpeechRecognizer(speechConfig, audioConfig);
+        const recognizer = multiLingualSpeechRecognizer(subscriptionKey, serviceRegion, [
+          "en-US",
+          "fr-FR",
+          "de-DE",
+          "it-IT"
+        ]);
```
superhindupur, line 175: This file is doing too much, and has many many imports. I needed to test the multi-lingual configuration for the speech recognizer, so I extracted it into another file, and tested that file. As we're adding new functionality to CWYD, we should try and modularize this code so that each file is of a manageable size and complexity and is conducive for testing.

superhindupur, line 175: Also, as mentioned on the PR description, the plan is to move this hardocded list of languages into the deployment config under #317 .

Makefile
```diff
@@ -38,9 +38,9 @@ uitest: ## 🧪 Run the ui tests in headless mode
 	@echo -e "\e[34m$@\e[0m" || true
 	@cd tests/integration/ui && npm install && npx cypress run --env ADMIN_WEBSITE_NAME=$(ADMIN_WEBSITE_NAME),FRONTEND_WEBSITE_NAME=$(FRONTEND_WEBSITE_NAME)
 
-build-frontend: ## 🏗️ Build the Frontend webapp
+unittest-frontend: ## 🏗️ Unit test the Frontend webapp
```
adamdougal, line 41: Did you mean to replace `build-frontend` with this? It might be worth having both `build` and `unittest` tasks for the front end.

code/frontend/src/util/SpeechToText.tsx
```diff
@@ -0,0 +1,22 @@
...
+    subscriptionKey: string,
+    serviceRegion: string,
+    languages: string[]
+  ) => {
+        const speechConfig = SpeechConfig.fromSubscription(
+          subscriptionKey,
+          serviceRegion
+        );
+        const audioConfig = AudioConfig.fromDefaultMicrophoneInput();
+        var autoDetectSourceLanguageConfig = AutoDetectSourceLanguageConfig.fromLanguages(languages)
```
adamdougal, line 19: Nit: Is there any reason this is a `var` and the rest are `const`? Probably doesn't matter either way, just noticed it was different!

adamdougal (changes requested):

Makefile
superhindupur: I replaced it because it followed the same pattern as we have for the python code - we don't have an explicit "build" command, we only have a unit-test. 

I thought the only reason we had the build-frontend was because there weren't any tests to run. Do you see any benefit in keeping the build-frontend?

code/frontend/src/util/SpeechToText.tsx
superhindupur: Oops, missed it when changing the others to const. Will edit this, thanks!

code/frontend/.npmignore
cecheta, line 1: Is this file required? I thought it was for when publishing an npm package

cecheta:
Could we get a ticket to include these tests in the GitHub Action?

Makefile
superhindupur: lol, my brain is super-wired to java. Ok, I don't feel very strongly on the subject so I don't mind reinstating the build-frontend command, but I will leave the ci command to call the unittest-frontend command directly.

code/frontend/.npmignore
superhindupur: You're right, I was assuming that this is what keeps the test files from being included in the final index.js asset created by npm run build - but no, it turns out to be the include: ["src"] setting in the tsconfig.json file (https://www.typescriptlang.org/tsconfig#include).

I removed the .npmigore file now. I double checked that the test file isn't included in npm run build by writing invalid code in it and npm run build still succeeded (whereas npm run test failed)

superhindupur:
> Could we get a ticket to include these tests in the GitHub Action?

@cecheta I added this as a subtask of #317

adamdougal (approved):
lgtm

cecheta (approved):
Nice 😄

PR #589, opened 2024-04-04
Functional tests for /azure_byod

cecheta:
## Purpose

* This PR adds functional tests for the `/api/conversation/azure_byod` endpoint, both when using data and when not using data.

In order to do this, it was necessary to launch the application a second time with different environment variables, to distinguish between using data and not using data. There is a refactor of the existing code to accommodate this, mainly that a  number of fixtures are changed to `package` scope, so they cleanup after the tests in that package have finished running. There is duplication between fixtures in the two packages, I wasn't sure how to avoid this while still ensuring that the fixtures cleaned up after running in each package.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x] Other... Please describe: Tests
```

## How to Test
* `poetry run pytest -m "not azure"


code/tests/functional/backend_api_without_data/__init__.py
adamdougal, line 1: I'm wondering if it would be cleaner/nicer to have sub dirs underneath `backend_api` e.g:

```
...
   backend_api/
      with_data/
      without_data/
```

Maybe with a small readme explaining the differences and where to put what tests. WDYT?

code/tests/functional/backend_api/test_azure_byod.py
```diff
@@ -0,0 +1,128 @@
...
+    app_url: str, app_config: AppConfig, httpserver: HTTPServer
+):
+    # given
+    httpserver.clear_all_handlers()  # Clear default successful responses
+
+    httpserver.expect_request(
+        f"/openai/deployments/{app_config.get('AZURE_OPENAI_MODEL')}/chat/completions",
+        method="POST",
+    ).respond_with_data(
+        Template(
```
adamdougal, line 36: Nice, didn't know you could do this!

code/tests/functional/backend_api/test_azure_byod.py
```diff
@@ -0,0 +1,128 @@
...
+                        "role": "assistant",
+                    },
+                ]
+            }
+        ],
+    }
+
+    assert response.headers["Content-Type"] == "application/json-lines"
+
+    verify_request_made(
```
adamdougal, line 83: This test is quite large, is it worth pulling this verification out into it's own test to break it up and reduce the number of reasons this test could fail? Same for the without data test.

code/tests/functional/backend_api_without_data/__init__.py
cecheta: I've moved some stuff around, what do you think?

adamdougal (approved):
lgtm

PR #619, opened 2024-04-08
Disable app insights when running locally

### LINKED ISSUE #634: Disable App Insights when running locally
# Motivation

We should disable app insights when running locally, so logs are sent to stdout and not Azure.

We already have an env var `APPINSIGHTS_ENABLED`, which we can use, however:
- It's not defined anywhere, but defaults to `True`
- The only place where it's used is using the old `AzureLogHandler` and should be updated for the new package
- The custom dimensions do not currently work

image

We cannot use `EnvHelper` as we need to configure Azure Monitor as soon as possible, before importing other libraries that would be instrumented.

# Requirements

- Disable call to `configure_azure_monitor()` when `APPINSIGHTS_ENABLED` is `False`
- Fix invalid dimensions in `TokenLogger.py`

cecheta:
Closes #634 

## Purpose

* Disable app insights when running locally, by using env var `APPINSIGHTS_ENABLED` which is only set in Azure
* Update `TokenLogger` to  not use deprecated opencensus library
* Fix error with custom dimensions
  * Note that the log message is not visible, as Flask sets the default log level to Warning

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
* Run webapp locally, see logs in stdout
* Run webapp in Azure, see logs in App Insights


cecheta:
> What if we check if the app insights connection string is present, rather than introducing a new env variable?

I guess that would work, but we'd need to remove these: https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/main/infra/main.bicep#L785-L787

@ross-p-smith do you see any downsides in removing these?

ross-p-smith:
It looks like the EnvHelper already has a APPINSIGHTS_ENABLED

>> @ross-p-smith do you see any downsides in removing these?

The reason we have duplicates is that the code relies on APPINSIGHTS_CONNECTION_STRING and I believe some libraries (I think one of the of aisearch ones) fails if APPLICATIONINSIGHTS_CONNECTION_STRING is not present. We should try and get rid of InstrumentationKey if possible. I think we can do as Tanya suggested if we keep it all hidden away in the EnvHelper (or whatever it becomes!)

cecheta:
> It looks like the EnvHelper already has a APPINSIGHTS_ENABLED
> 
> > > @ross-p-smith do you see any downsides in removing these?
> 
> The reason we have duplicates is that the code relies on APPINSIGHTS_CONNECTION_STRING and I believe some libraries (I think one of the of aisearch ones) fails if APPLICATIONINSIGHTS_CONNECTION_STRING is not present. We should try and get rid of InstrumentationKey if possible. I think we can do as Tanya suggested if we keep it all hidden away in the EnvHelper (or whatever it becomes!)

I'm a bit hesitant to use the EnvHelper as I think we should be calling `configure_azure_monitor()` as soon as possible, before we start importing other packages, like in this example: https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/monitor/azure-monitor-opentelemetry/samples/tracing/azure_core.py. Otherwise it might not work correctly, as we saw with Flask: https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/597

What I meant by removing those outputs was so that those values would not be set within the env vars that come from `azd`. Then we would do something like:

```py
if os.getenv("APPLICATIONINSIGHTS_CONNECTION_STRING") or os.getenv("APPINSIGHTS_INSTRUMENTATIONKEY"):
    configure_azure_monitor()
```

code/app.py
```diff
@@ -1,8 +1,10 @@
+import os
 from azure.monitor.opentelemetry import configure_azure_monitor
 from opentelemetry.instrumentation.httpx import HTTPXClientInstrumentor
 
-configure_azure_monitor()
-HTTPXClientInstrumentor().instrument()  # httpx is used by openai
+if os.getenv("APPLICATIONINSIGHTS_ENABLED", "false").lower() == "true":
```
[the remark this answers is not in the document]
cecheta: Updated to `APPINSIGHTS_ENABLED`

code/backend/batch/utilities/loggers/TokenLogger.py
```diff
@@ -1,24 +1,15 @@
 import logging
-from opencensus.ext.azure.log_exporter import AzureLogHandler
-from ..helpers.EnvHelper import EnvHelper
 
 
+# TODO: We probably don't need a separate class for this
```
ross-p-smith, line 4: Can we raise another issue for this and I think this is the only code that needs APPINSIGHTS_CONNECTION_STRING - so can we get rid of that too? Happy for you to close this comment when we have a linked issue to clean these connection strings up (unless you have one already!)

code/backend/batch/utilities/helpers/EnvHelper.py
```diff
@@ -128,7 +128,7 @@ def __init__(self, **kwargs) -> None:
             "AZURE_FORM_RECOGNIZER_KEY"
         )
         # Azure App Insights
-        self.APPINSIGHTS_ENABLED = self.get_env_var_bool("APPINSIGHTS_ENABLED")
+        self.APPINSIGHTS_ENABLED = self.get_env_var_bool("APPINSIGHTS_ENABLED", "False")
```
ross-p-smith, line 131: Is it worth adding a comment here to say that this will be turned on within an AppService when an App Insights is present?

ross-p-smith (approved):

code/backend/batch/utilities/helpers/EnvHelper.py
cecheta: Added a comment, is that what you were looking for?

code/backend/batch/utilities/loggers/TokenLogger.py
cecheta: Raised #637 and #638

ross-p-smith (approved):

PR #626, opened 2024-04-08
Use `context` in 'Find associated pull request' script

cecheta:
## Purpose

* Use `context` object in GitHub script rather than directly from `github` object.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
Difficult to test, as it only runs from `main`, however I have tested it in my fork.

image

### DIFF
--- a/.github/workflows/comment_coverage.yml
+++ b/.github/workflows/comment_coverage.yml
@@ -31,9 +31,9 @@ jobs:
           script: |
             const response = await github.rest.pulls.list({
               owner: "${{ github.repository_owner }}",
-              repo: "${{ github.event.workflow_run.repository.name }}",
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -24,7 +24,6 @@ jobs:
           python-version: "3.11"
           architecture: x64
           cache: pip
-          cache-dependency-path: "**/pyproject.toml"
```
cecheta, line 27: The cache should be dependent on `poetry.lock`, which is the default

ross-p-smith (approved):

PR #632, opened 2024-04-08
CWE-20 - Code Security

ross-p-smith:
## Purpose
When checking for Blob Storage we should only check that the domain ends with blob storage domain rather than it just contains it based on Code Security CWE-20

## How to Test
Tested by Unit Tests

### DIFF
--- a/code/backend/batch/utilities/common/SourceDocument.py
+++ b/code/backend/batch/utilities/common/SourceDocument.py
@@ -61,7 +61,8 @@ class SourceDocument:
         hash_key = f"doc_{hash_key}"
         sas_placeholder = (
             "_SAS_TOKEN_PLACEHOLDER_"
-            if "blob.core.windows.net" in parsed_url.netloc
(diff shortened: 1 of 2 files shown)


code/backend/batch/utilities/common/SourceDocument.py
```diff
@@ -61,7 +61,8 @@ def from_metadata(
         hash_key = f"doc_{hash_key}"
         sas_placeholder = (
             "_SAS_TOKEN_PLACEHOLDER_"
-            if "blob.core.windows.net" in parsed_url.netloc
+            if parsed_url.netloc
+            and parsed_url.netloc.endswith(".blob.core.windows.net")
```
adamdougal, line 65: Is it worth adding a test which exercises this?

ross-p-smith: Yep sorry - done!

adamdougal (approved):
LGTM!

PR #644, opened 2024-04-09
Add test for backend functions

tanya-borisova:
## Purpose

* Adding unit tests for the backend functions.

```
[ ] Yes
[ x ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ x ] Other... Please describe: **Adding tests**
```


code/tests/test_BatchGetConversationResponse.py
cecheta, line 1: Nit: this file should be `test_GetConversationResponse.py`

code/tests/test_AddURLEmbeddings.py
cecheta, line 1: I think it would be good to do a test when there's no url in the request

code/backend/batch/AddURLEmbeddings.py
```diff
@@ -1,20 +1,21 @@
...
-sys.path.append("..")
 
 bp_add_url_embeddings = func.Blueprint()
-
 logger = logging.getLogger(__name__)
 
 
 @bp_add_url_embeddings.route(route="AddURLEmbeddings")
 def add_url_embeddings(req: func.HttpRequest) -> func.HttpResponse:
+    do_add_url_embeddings(req)
```
cecheta, line 15:
```suggestion
    return do_add_url_embeddings(req)
```

code/backend/batch/GetConversationResponse.py
```diff
@@ -1,20 +1,23 @@
...
 
 bp_get_conversation_response = func.Blueprint()
 env_helper: EnvHelper = EnvHelper()
-
 logger = logging.getLogger(__name__)
 
 
 @bp_get_conversation_response.route(route="GetConversationResponse")
 def get_conversation_response(req: func.HttpRequest) -> func.HttpResponse:
+    do_get_conversation_response(req)
```
cecheta, line 17:
```suggestion
    return do_get_conversation_response(req)
```

cecheta (changes requested):

cecheta:
The tests are quite slow, I think we need to mock `ConfigHelper` in more of the tests

code/backend/batch/AddURLEmbeddings.py
```diff
@@ -1,20 +1,21 @@
...
-sys.path.append("..")
 
 bp_add_url_embeddings = func.Blueprint()
-
 logger = logging.getLogger(__name__)
 
 
 @bp_add_url_embeddings.route(route="AddURLEmbeddings")
 def add_url_embeddings(req: func.HttpRequest) -> func.HttpResponse:
+    return do_add_url_embeddings(req)
```
adamdougal, line 15: Why is this required? Can't you just import `add_url_embeddings` in the tests?

code/backend/batch/AddURLEmbeddings.py
```diff
@@ -1,20 +1,21 @@
 import logging
 import traceback
 import azure.functions as func
-import sys
 
 from utilities.helpers.DocumentProcessorHelper import DocumentProcessor
 from utilities.helpers.ConfigHelper import ConfigHelper
 
-sys.path.append("..")
```
adamdougal, line 9: Do we know why this was needed? Do the deployed functions still work without this?

code/backend/batch/AddURLEmbeddings.py
```diff
@@ -1,20 +1,21 @@
 import logging
 import traceback
 import azure.functions as func
-import sys
 
 from utilities.helpers.DocumentProcessorHelper import DocumentProcessor
 from utilities.helpers.ConfigHelper import ConfigHelper
```
adamdougal, line 6: If I change this line, and all the other imports to be prepended with `backend.batch.` to:
```suggestion
from backend.batch.utilities.helpers.ConfigHelper import ConfigHelper
```

I don't need this in the test files:
```python
function_app_path = os.path.abspath(
    os.path.join(os.path.dirname(__file__), "../backend/batch")
)
sys.path.append(function_app_path)
```

Does that work for you?

cecheta: I was also going to suggest this, but then I realised that when this gets deployed, the `batch` folder becomes the root, so it doesn't know about `backend.batch` at runtime

adamdougal: Gotcha, thanks!

code/backend/batch/AddURLEmbeddings.py
tanya-borisova: No, it doesn't work, I suspect because of the decorator

adamdougal: Ah I think you can make it work by first "building" the function. I just followed this guide and it worked for one of the tests https://medium.com/mesh-ai-technology-and-engineering/writing-and-testing-azure-functions-in-the-v2-python-programming-model-c391bd779ff6

tanya-borisova: By "doesn't work" I mean that the function call simply won't execute from the test, you can try to change this and set the breakpoint inside add_url_embeddings when running the test - it won't stop there

adamdougal: Is that even following the guide above?

I only tried it with one test, but the below worked for me:
```python
@patch("backend.batch.AddURLEmbeddings.ConfigHelper")
@patch("backend.batch.AddURLEmbeddings.DocumentProcessor")
def test_add_url_embeddings_when_url_set_in_body(_, __):
    fake_request = func.HttpRequest(
        method="POST",
        url="",
        body=b'{"url": "https://example.com"}',
        headers={"Content-Type": "application/json"},
    )

    func_call = add_url_embeddings.build().get_user_function() # Note this line <----------
    response = func_call(fake_request)

    assert response.status_code == 200
```

code/backend/batch/AddURLEmbeddings.py
tanya-borisova: I don't know why this was added. (My guess would be that there was some refactoring and this was left behind). The deployed functions still work, verified both in the deployment and run locally with `poetry func start`. 

I also think it's much more robust to use an absolute path with sys.path.append as I do here in the test files

code/backend/batch/AddURLEmbeddings.py
tanya-borisova: Sorry I hadn't seen your reply when I typed in the comment, will try this

tanya-borisova: The method is now triggered but weirdly all paths for `patch` call have broken now

code/backend/batch/AddURLEmbeddings.py
```diff
@@ -1,24 +1,26 @@
...
-logger.setLevel(env_helper.LOGLEVEL)
 
 
 @bp_add_url_embeddings.route(route="AddURLEmbeddings")
 def add_url_embeddings(req: func.HttpRequest) -> func.HttpResponse:
+    return do_add_url_embeddings(req)
+
+
+def do_add_url_embeddings(req: func.HttpRequest) -> func.HttpResponse:
+    logger = logging.getLogger(__name__)
```
cecheta, line 19: Is it possible to still leave the configuration of the logger where it was before?

code/tests/test_BatchPushResults.py
```diff
@@ -0,0 +1,69 @@
+import sys
+import os
+from unittest.mock import patch, Mock
+from azure.functions import QueueMessage
+
+
+sys.path.append(os.path.join(os.path.dirname(sys.path[0]), "backend", "batch"))
+
+from backend.batch.BatchPushResults import do_batch_push_results  # noqa: E402
+from backend.batch.BatchPushResults import _get_file_name_from_message  # noqa: E402
```
cecheta, line 10: Nit: Could you combine these two imports?

code/backend/batch/AddURLEmbeddings.py
tanya-borisova: I think I can move create the logger itself outside, but to set the log level it now depends on the EnvHelper which needs to be mocked

code/backend/batch/AddURLEmbeddings.py
tanya-borisova: ^ looks like it was my fault for breaking the patch, this has been applied now so I'm going to resolve this thread

code/backend/batch/AddURLEmbeddings.py
cecheta: Oh I see

I think we should get it from `os.environ` without using the helper, like here

https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/f39308437acbcbaa51f0ab57ed9ba8e34d695893/code/app.py#L6

tanya-borisova: Tbh I'm not a fan of this, imo it would be better if the EnvHelper didn't come with so many side effects, but that's beside the scope of this PR :)

adamdougal (approved):
Thanks for this!

code/backend/batch/AddURLEmbeddings.py
cecheta: Same tbh

cecheta (approved):
Thank you!!

PR #646, opened 2024-04-10
Stop B3 being downsized to B1

ross-p-smith:
## Purpose
Stop B3 being downsized to B1

## How to Test
Wait a day!


adamdougal:
Can you explain how this fixes the problem? Thanks!

PR #653, opened 2024-04-11
Allow openai model to be configured via azd

adamdougal:
Raising off the back of https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/discussions/431

### DIFF
--- a/infra/main.bicepparam
+++ b/infra/main.bicepparam
@@ -13,6 +13,8 @@ param authType = readEnvironmentVariable('AZURE_AUTH_TYPE', 'keys')
 
 param hostingModel = readEnvironmentVariable('AZURE_APP_SERVICE_HOSTING_MODEL', 'code')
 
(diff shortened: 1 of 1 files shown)


infra/main.bicepparam
```diff
@@ -13,6 +13,8 @@ param authType = readEnvironmentVariable('AZURE_AUTH_TYPE', 'keys')
 
 param hostingModel = readEnvironmentVariable('AZURE_APP_SERVICE_HOSTING_MODEL', 'code')
 
+param azureOpenAIModel = readEnvironmentVariable('AZURE_OPENAI_MODEL', 'gpt-4')
```
ross-p-smith, line 16: This would mean that if we deploy from "Deploy to Azure" the default is 3.5; but deploying from `azd up` it would be 4.

Do you think it's best to leave as 3.5 until all regions support 4?

adamdougal: Oops, my bad. I'll lower

cecheta (approved):

Issue #517, opened 2024-03-22
Not able to implement hybrid search(semantic and vector). Only vector search is present. I believe because of this, the chat is not full functional.


PFA23SCM89S:
# Describe the bug

# Expected behavior

# How does this bug make you feel?
_Share a gif from [giphy](https://giphy.com/) to tells us how you'd feel_

---

# Debugging information

## Logs

If applicable, add logs to help the engineer debug the problem.

---

# Tasks

_To be filled in by the engineer picking up the issue_

- [ ] Task 1
- [ ] Task 2
- [ ] ...

cecheta:
Hi @PFA23SCM89S , thank you for raising this bug. Please could you provide us with the steps to reproduce this bug and what we should expect to see vs what actually happens? Thank you.

cherifbenham:
question answer tool is not dynamic regarding search and top k

![image](https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/assets/99403039/b9571393-3832-4e2a-882d-78cd0f28e7c4)

PFA23SCM89S:
@cecheta To reproduce this error, just upload any document in admin webpage and ask specific questions from the document, preferably a bigger prompt. Not sure how to fix this.
@cherifbenham Any idea on what to change in the code?

cherifbenham:
@PFA23SCM89S in your case, if you deployed from devcontainer without changing the code, then your search should be hybrid meaning text+vector that outputs the 4 closest docs (or less)

the bug i am trying to fix relates to the implementation of semantic+hybrid, i found a way to do it using semantic config 

semantic_config = SemanticConfiguration(name="semantic_config", prioritized_fields=SemanticPrioritizedFields(title_field=SemanticField(field_name="title"),content_fields=["content"], keywords_fields=["title"]")

and then

sources = self.vector_store.similarity_search(query=question, k=10, search_type = "semantic_hybrid", semanticConfiguration="semantic_config")

i am trying to redeploy like this - will let you know how it goes

PFA23SCM89S:
@cherifbenham Thanks for the update. I tried to update my open ai version from gpt 35 turbo to gpt 4. Still the response i get is very poor. im not sure if the model is updated provide. can u help me with how to update the model properly.

cherifbenham:
You can try to update the model version from gpt35 to gpt4 in env variables
of function apps and in environment variable of web app.
[...]

Issue #295, opened 2024-02-16
Allow for different queryType settings for chat on your data to Azure AI Search 


vermegi:
# Motivation

Currently search results on your data are not always performing as expected. It would be nice to be able to switch and test out different query types to Azure AI Search, currently the code is fixed on semantic search and this does not always give good chat results: https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/f7dd41ed4def1007465c526320fe9eb9173ecbee/code/app/app.py#L132  
Different queryType settings are documented at https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#azure-ai-search-parameters 
Since the Index in Azure AI Search contains vector, also allow for the vector query types for better results. 

# Requirements

- Enable the configuration of vector and hyrbrid query types

# Tasks

_To be filled in by the engineer picking up the issue_

- [ ] Task 1
- [ ] Task 2
- [ ] ...

> ---------------------------------------------------------------
> Thanks! We'll be in touch soon.

PR #654, opened 2024-04-11
Enable semantic search + filtering

### LINKED ISSUE #449: Enable Semantic Search - after deployment from devcontainer
# Motivation

It seems like adding a semantic configuration in the Search Index (documents) does not impact the chat/web experience.
I created new Env Variables in Functions Apps resource: AZURE_SEARCH_USE_SEMANTIC_SEARCH=true and AZURE_SEARCH_SEMANTIC_SEARCH_CONFIG=default where there is a semantic configuration in search named 'default'
But search remains as in deployment initial conditions. Please advise

Amazing

# Requirements

- Search functions are not impacted by a change in TOP_K or by SEMANTIC_CONFIG

# Tasks

- Update Search Instances so that they take into account TOP_K and SEMANTIC_CONFIG
- Ideally, add a filter parameter in search

cecheta:
Closes #449

## Purpose

This PR:
* Adds support for `AZURE_SEARCH_TOP_K` and `AZURE_SEARCH_DIMENSIONS` when using custom endpoint
* Adds new env var `AZURE_SEARCH_FILTER` to both endpoints, for filtering search results
* Adds `poetry install` before deploying app
* Remove unnecessary call to OpenAI embeddings just to get `1536`
* Update tests

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
* Run code locally
* Configure the following env vars
  * AZURE_SEARCH_TOP_K
  * AZURE_SEARCH_FILTER
  * AZURE_SEARCH_DIMENSIONS
  * AZURE_SEARCH_USE_SEMANTIC_SEARCH
  * AZURE_SEARCH_SEMANTIC_SEARCH_CONFIG

## What to Check
* Make requests to `/azure_byod` and `/custom` endpoints
* Verify that requests are as expected, using semantic search + filtering as appropriate

### DIFF
--- a/code/backend/batch/utilities/helpers/AzureSearchHelper.py
+++ b/code/backend/batch/utilities/helpers/AzureSearchHelper.py
@@ -10,12 +10,21 @@ from .EnvHelper import EnvHelper
 
 
 class AzureSearchHelper:
+    _search_dimension: int | None = None
+
     def __init__(self):
-        pass
+        self.llm_helper = LLMHelper()
+        self.env_helper = EnvHelper()
+
+    @property
+    def search_dimensions(self) -> int:
+        if AzureSearchHelper._search_dimension is None:
+            AzureSearchHelper._search_dimension = len(
+                self.llm_helper.get_embedding_model().embed_query("Text")
+            )
+        return AzureSearchHelper._search_dimension
 
     def get_vector_store(self):
-        llm_helper = LLMHelper()
-        env_helper = EnvHelper()
         fields = [
             SimpleField(
                 name="id",
@@ -31,7 +40,7 @@ class AzureSearchHelper:
                 name="content_vector",
                 type=SearchFieldDataType.Collection(SearchFieldDataType.Single),
                 searchable=True,
-                vector_search_dimensions=int(env_helper.AZURE_SEARCH_DIMENSIONS),
+                vector_search_dimensions=self.search_dimensions,
                 vector_search_profile_name="myHnswProfile",
             ),
             SearchableField(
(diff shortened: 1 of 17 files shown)


ross-p-smith:
Fantastic! Can you have a word with @reechamishra as I think she was trying to pull together all the issues that are related to this

code/backend/batch/utilities/helpers/AzureSearchHelper.py
```diff
@@ -31,9 +31,7 @@ def get_vector_store(self):
                 name="content_vector",
                 type=SearchFieldDataType.Collection(SearchFieldDataType.Single),
                 searchable=True,
-                vector_search_dimensions=len(
-                    llm_helper.get_embedding_model().embed_query("Text")
-                ),
+                vector_search_dimensions=int(env_helper.AZURE_SEARCH_DIMENSIONS),
```
ross-p-smith, line 34: Can we leave a link here to where people will find what this value should be?
https://learn.microsoft.com/en-us/azure/search/vector-search-index-size?tabs=rest-vector-quota#memory-overhead-from-the-selected-algorithm

(and maybe in the EnvHelper)

code/backend/batch/utilities/helpers/AzureSearchHelper.py
```diff
@@ -31,9 +31,7 @@ def get_vector_store(self):
                 name="content_vector",
                 type=SearchFieldDataType.Collection(SearchFieldDataType.Single),
                 searchable=True,
-                vector_search_dimensions=len(
-                    llm_helper.get_embedding_model().embed_query("Text")
```
ross-p-smith, line 35: I understand that we don't need to call `llm_helper.get_embedding_model().embed_query` if we already have that value in config. But we make 3 calls to that in this method anyway. Is it worth calling it once at the top anyway.

Would the call to that method every not return what was configured in AZURE_SEARCH_DIMENSIONS?

ross-p-smith (dismissed):
💖💖💖💖

ross-p-smith:
@cecheta - can you go through these and see which ones you can set to be closed (if relevant) #295, #449, #517

code/backend/batch/utilities/helpers/AzureSearchHelper.py
cecheta: We don't need to call this method at all, essentially what was happening before was that we were getting the embeddings for a dummy string, i.e. "Text", and using the length of the array of embeddings to calculate what the dimension size should be. But for this model, `text-embedding-ada-002` it's always `1536`, so we don't need to make a call to OpenAI, we can just use the value directly, as set in the env var.

https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/models#embeddings-models

Also, in the other places where `llm_helper.get_embedding_model().embed_query` is still present, we're not actually calling the function, we're just passing the function to `AzureSearch`

code/backend/batch/utilities/helpers/AzureSearchHelper.py
cecheta: Wouldn't the README make more sense?

Also I think it would be better to link to this: https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/models#embeddings-models

ross-p-smith: Yes - much better! :-)

code/backend/batch/utilities/helpers/AzureSearchHelper.py
adamdougal: What if a different embeddings model is used, will this change break that?

ross-p-smith (approved):

adamdougal (approved):
LGTM!

code/backend/batch/utilities/helpers/AzureSearchHelper.py
cecheta: If a different model would be used then they would have to change the value of the env var, I don't think that would be the worst thing in the world, however I've updated it now so that we call the embeddings once and then cache it in the class, so we don't need to calculate it again.

I'm not sure how much we're really saving here, since everytime we call `AzureSearch()` from langchain, that module does the exact same thing and gets the embeddings for `"Text"`, however I think the caching in the class (or something similar) is a pattern that we should look to put in multiple places to reduce latency in the app.

ross-p-smith (approved):

PR #660, opened 2024-04-12
Make the speech recognizer's supported languages configurable. Required by #317

superhindupur:
## Purpose
This PR removes the previously hardcoded list of spoken languages recognized by the chat web interface and reads the list from the deployment configuration instead. Required by #317 

The parameter is accepted by the deployment bicep script, and is outputted to the azure env file. It is then injected at build time  to the frontend UI react app via an environment variable. 

At runtime, this environment variable is read by the frontend UI app to configure the Azure Speech recognizer to recognize the given languages via speech. 

```
[ ] Yes
[x ] No
```

## Pull Request Type

```
[ ] Bugfix
[x ] Feature
[ ] Other... Please describe:
```

## How to Test
Deploy the code, then test via the web chat app that the configured languages are being recognized by the speech input.
I tested English and Italian at my end.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,10 @@ uitest: ## 🧪 Run the ui tests in headless mode
 	@echo -e "\e[34m$@\e[0m" || true
 	@cd tests/integration/ui && npm install && npx cypress run --env ADMIN_WEBSITE_NAME=$(ADMIN_WEBSITE_NAME),FRONTEND_WEBSITE_NAME=$(FRONTEND_WEBSITE_NAME)
 
+build-frontend: ## 🏗️ Build the Frontend webapp
+	@echo -e "\e[34m$@\e[0m" || true
+	@cd code/frontend && npm install && npm run build
+
 unittest-frontend: ## 🏗️ Unit test the Frontend webapp
 	@echo -e "\e[34m$@\e[0m" || true
 	@cd code/frontend && npm install && npm run build && npm run test
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ SHELL := /bin/bash
 .DEFAULT_GOAL := help
 
 AZURE_ENV_FILE := $(shell azd env list --output json | jq -r '.[] | select(.IsDefault == true) | .DotEnvPath')
+AZURE_ENV_DIR := $(shell dirname $(AZURE_ENV_FILE))
 
 ENV_FILE := .env
 ifeq ($(filter $(MAKECMDGOALS),config clean),)
@@ -40,11 +41,11 @@ uitest: ## 🧪 Run the ui tests in headless mode
 
 build-frontend: ## 🏗️ Build the Frontend webapp
 	@echo -e "\e[34m$@\e[0m" || true
-	@cd code/frontend && npm install && npm run build
+	@cd code/frontend && npm install && VITE_ENV_DIR=$(AZURE_ENV_DIR) npm run build
 
-unittest-frontend: ## 🏗️ Unit test the Frontend webapp
+unittest-frontend: build-frontend ## 🏗️ Unit test the Frontend webapp
 	@echo -e "\e[34m$@\e[0m" || true
-	@cd code/frontend && npm install && npm run build && npm run test
+	@cd code/frontend && npm run test
 
 unittest-frontend: ## 🏗️ Unit test the Frontend webapp
 	@echo -e "\e[34m$@\e[0m" || true
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,6 @@ SHELL := /bin/bash
 .DEFAULT_GOAL := help
 
 AZURE_ENV_FILE := $(shell azd env list --output json | jq -r '.[] | select(.IsDefault == true) | .DotEnvPath')
-AZURE_ENV_DIR := $(shell dirname $(AZURE_ENV_FILE))
 
 ENV_FILE := .env
 ifeq ($(filter $(MAKECMDGOALS),config clean),)
@@ -41,15 +40,11 @@ uitest: ## 🧪 Run the ui tests in headless mode
 
 build-frontend: ## 🏗️ Build the Frontend webapp
 	@echo -e "\e[34m$@\e[0m" || true
-	@cd code/frontend && npm install && VITE_ENV_DIR=$(AZURE_ENV_DIR) npm run build
+	@cd code/frontend && npm install && npm run build
 
 unittest-frontend: build-frontend ## 🏗️ Unit test the Frontend webapp
 	@echo -e "\e[34m$@\e[0m" || true
 	@cd code/frontend && npm run test
-
-unittest-frontend: ## 🏗️ Unit test the Frontend webapp
-	@echo -e "\e[34m$@\e[0m" || true
-	@cd code/frontend && npm install && npm run build && npm run test
 
 azd-login: ## 🔑 Login to Azure with azd and a SPN
 	@echo -e "\e[34m$@\e[0m" || true
--- a/Makefile
+++ b/Makefile
@@ -38,9 +38,9 @@ uitest: ## 🧪 Run the ui tests in headless mode
 	@echo -e "\e[34m$@\e[0m" || true
 	@cd tests/integration/ui && npm install && npx cypress run --env ADMIN_WEBSITE_NAME=$(ADMIN_WEBSITE_NAME),FRONTEND_WEBSITE_NAME=$(FRONTEND_WEBSITE_NAME)
 
-build-frontend: ## 🏗️ Build the Frontend webapp
+unittest-frontend: ## 🏗️ Unit test the Frontend webapp
 	@echo -e "\e[34m$@\e[0m" || true
-	@cd code/frontend && npm install && npm run build
+	@cd code/frontend && npm install && npm run build && npm run test
 
 unittest-frontend: ## 🏗️ Unit test the Frontend webapp
 	@echo -e "\e[34m$@\e[0m" || true
--- a/code/backend/batch/utilities/helpers/EnvHelper.py
+++ b/code/backend/batch/utilities/helpers/EnvHelper.py
@@ -175,6 +175,9 @@ class EnvHelper:
         # Speech Service
         self.AZURE_SPEECH_SERVICE_NAME = os.getenv("AZURE_SPEECH_SERVICE_NAME", "")
         self.AZURE_SPEECH_SERVICE_REGION = os.getenv("AZURE_SPEECH_SERVICE_REGION")
+        self.SPEECH_RECOGNIZER_LANGUAGES = self.get_env_var_array(
+            "SPEECH_RECOGNIZER_LANGUAGES"
+        )
         self.AZURE_SPEECH_REGION_ENDPOINT = os.environ.get(
             "AZURE_SPEECH_REGION_ENDPOINT",
             f"https://{self.AZURE_SPEECH_SERVICE_REGION}.api.cognitive.microsoft.com/",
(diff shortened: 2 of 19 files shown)


code/frontend/package.json
```diff
@@ -7,7 +7,7 @@
         "dev": "tsc && vite",
         "build": "tsc && vite build",
         "watch": "tsc && vite build --watch",
-        "test": "mocha"
+        "test": "vitest run"
```
superhindupur, line 10: I realized that vitest is much more lightweight as a unit test framework when testing in typescript. Plus it allows for super-easy mocking of environment variables.

code/frontend/package.json
```diff
@@ -30,19 +30,15 @@
...
         "@types/uuid": "^9.0.8",
         "@vitejs/plugin-react": "^4.2.1",
-        "chai": "^5.1.0",
-        "mocha": "^10.4.0",
         "prettier": "^3.2.5",
-        "ts-node": "^10.9.2",
         "typescript": "^5.4.4",
-        "vite": "^5.2.8"
+        "vite": "^5.2.8",
+        "vitest": "^1.5.0"
```
superhindupur, line 42: I was able to get rid of a bunch of dependencies now that we're using vitest and not mocha.

scripts/package_frontend.ps1
```diff
@@ -10,4 +10,4 @@ cp backend dist -r -Force
 # Node
 cd frontend
 npm install
-npm run build
+VITE_ENV_DIR=$(dirname $(azd env list --output json | jq -r '.[] | select(.IsDefault == true) | .DotEnvPath')) npm run build
```
superhindupur, line 13: I'm not sure how to test this. @ross-p-smith did you actually deploy from windows (not WSL) to test this flow with the powershell script and deployment?

scripts/package_frontend.sh
```diff
@@ -8,6 +8,7 @@ poetry export -o dist/requirements.txt
 cp *.py dist
 cp -r backend dist
 
+
 cd frontend
 npm install
-npm run build
+VITE_ENV_DIR=$(dirname $(azd env list --output json | jq -r '.[] | select(.IsDefault == true) | .DotEnvPath')) npm run build
```
superhindupur, line 14: This, I've tested successfully

code/frontend/vite.config.ts
```diff
@@ -1,8 +1,11 @@
 import { defineConfig } from "vite";
 import react from "@vitejs/plugin-react";
 
+const envDir = process.env.VITE_ENV_DIR;
```
ross-p-smith, line 4: Apologies if I misunderstand the code here. Are we now tying the typescript app in to having to have an .env file read in rather than having the hosting process deal with that? At the moment whether the code runs in App Service or in a Container or locally; the host/process has environment variables setup before the app runs - does this mean you know have to copy an .env in at build time?

scripts/package_frontend.ps1
ross-p-smith: Yes I set everything up in Windows and tested. Why does the build need to know where the env file is? Isnt that a runtime concern?

code/frontend/test/SpeechToText.test.ts
```diff
@@ -1,16 +1,22 @@
...
+
+
+    const recognizer = multiLingualSpeechRecognizer(key, region);
 
     expect(recognizer.properties.getProperty("SpeechServiceConnection_Key")).to.equal(key);
     expect(recognizer.properties.getProperty("SpeechServiceConnection_Region")).to.equal(region);
-    expect(recognizer.properties.getProperty("SpeechServiceConnection_AutoDetectSourceLanguages")).to.equal(languages.join(","));
+    expect(recognizer.properties.getProperty("SpeechServiceConnection_AutoDetectSourceLanguages")).to.equal(recognizedLanguages);
   });
 });
```
adamdougal, line 22: Is it worth adding a test for the defaulting of the languages if the env var is not set?

infra/main.bicep
```diff
@@ -189,6 +189,9 @@ param hostingModel string = 'container'
 ])
 param logLevel string = 'INFO'
 
+@description('List of comma-separated languages to recognize from the speech input (e.g. en-US,es-ES).')
```
adamdougal, line 192: Is there a defined list of these language codes that we can reference?

infra/main.bicep
```diff
@@ -866,4 +869,5 @@ output USE_KEY_VAULT bool = useKeyVault
 output AZURE_APP_SERVICE_HOSTING_MODEL string = hostingModel
 output FRONTEND_WEBSITE_NAME string = hostingModel == 'code' ? web.outputs.FRONTEND_API_URI : web_docker.outputs.FRONTEND_API_URI
 output ADMIN_WEBSITE_NAME string = hostingModel == 'code' ? adminweb.outputs.WEBSITE_ADMIN_URI : adminweb_docker.outputs.WEBSITE_ADMIN_URI
+output VITE_SPEECH_RECOGNIZER_LANGUAGES string = recognizedLanguages
```
adamdougal, line 872: Does this also need to be added to the app settings for the webapp? https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/main/infra/main.bicep#L372

ross-p-smith:
nit: Can you change the title of this PR and link it to an issue/bug/etc?

code/frontend/vite.config.ts
superhindupur: We are injecting a configuration into the typescript app at build time so that it is embedded into the frontend app when it is required. 

Now, we could also just have the frontend API layer read the env and return it via an API endpoint - that is another approach. I did think about that approached and picked this one since it would mean one less API call. I'm happy to reconsider if you think my current approach is making build-time more complicated. Thoughts?

scripts/package_frontend.ps1
ross-p-smith: But "why" do we need environment variables at build time? The values can change at runtime, so I don't get it. Having variables in a file is fine for local debugging etc, but I would want to be able to override them at run time, otherwise I lose the benefit of having it configurable.

Makefile
```diff
@@ -40,7 +41,11 @@ uitest: ## 🧪 Run the ui tests in headless mode
 
 build-frontend: ## 🏗️ Build the Frontend webapp
 	@echo -e "\e[34m$@\e[0m" || true
-	@cd code/frontend && npm install && npm run build
+	@cd code/frontend && npm install && VITE_ENV_DIR=$(AZURE_ENV_DIR) npm run build
+
+unittest-frontend: build-frontend ## 🏗️ Unit test the Frontend webapp
```
ross-p-smith, line 46: We don't need 2 `unittest-frontend` targets

ross-p-smith:
From my understanding, this change is all about passing in a string "en-US,fr-FR,de-DE,it-IT" to TypeScript.

1) I agree with the bicep changes to get this into an .env file for local debugging, but then as Adam states, it should be added to the AppSettings for the host "web" and "web.docker"
2) vite will read whatever is on process.env, so I don't see why it needs the path of the env file, rather as it should be able to read VITE_SPEECH_RECOGNIZER_LANGUAGES if it's loaded onto the process.
3) I would also question why the variable needs to be prefixed with VITE_ as that does not relate to it
4) I think you have gone down this path because it is not working in the makefile - and that is because we did not implement a previous change that I requested. I think you need something like 

```
AZURE_ENV_FILE := $(shell azd env list --output json | jq -r '.[] | select(.IsDefault == true) | .DotEnvPath')
ifeq ($(filter $(MAKECMDGOALS),config clean),)
	ifneq ($(strip $(wildcard $(AZURE_ENV_FILE))),)
		ifneq ($(MAKECMDGOALS),config)
			include $(AZURE_ENV_FILE)
			export
		endif
	endif
endif
```

This would actually "export" the file and the process would have the all of the environment variables loaded and vite would get your string. In the case of running in the cloud, then it would use the AppSettings

ross-p-smith (changes requested):
Left comment - I feel that we do not need to bring in a dependency on a file

scripts/package_frontend.ps1
cecheta: The frontend React app running in the background cannot read environment variables, I think we have to set them at build time, or fetch them from an API call

You could maybe piggy-back off this PR: #680

Makefile
superhindupur: That was by mistake. I will remove the old one

code/frontend/test/SpeechToText.test.ts
```diff
@@ -1,16 +1,36 @@
...
-  it("creates a speech recognizer with multiple languages", async () => {
+  beforeEach(() => {
+
+  })
+
+  it("creates a speech recognizer with the default language if config not specified", async () => {;
     const token = "token";
     const region = "region";
     const languages = ["en-US", "fr-FR", "de-DE", "it-IT"];
     const recognizer = multiLingualSpeechRecognizer(token, region, languages);
```
adamdougal, line 15: Is this being overridden by the line below?

superhindupur: I'd marked the PR as draft since I was in the middle of making changes. Will submit the PR for review soon.

infra/main.bicep
```diff
@@ -413,6 +416,7 @@ module web './app/web.bicep' = if (hostingModel == 'code') {
       AZURE_SEARCH_URL_COLUMN: azureSearchUrlColumn
       AZURE_SPEECH_SERVICE_NAME: speechServiceName
       AZURE_SPEECH_SERVICE_REGION: location
+      SPEECH_RECOGNIZER_LANGUAGES: recognizedLanguages
```
ross-p-smith, line 419: We will need `SPEECH_RECOGNIZER_LANGUAGES: recognizedLanguages` in the container section too. Just an idea, but you could move the param and the app setting to this file /workspaces/chat-with-your-data-solution-accelerator/infra/app/web.bicep which will do it for code and container. I suppose the decision is, do you think many people will change this per deployment - again - I'm happy either way

ross-p-smith:
Many thanks for this, one tiny addition for me and we can get this home!

code/backend/batch/utilities/helpers/EnvHelper.py
```diff
@@ -175,6 +175,9 @@ def __load_config(self, **kwargs) -> None:
         # Speech Service
         self.AZURE_SPEECH_SERVICE_NAME = os.getenv("AZURE_SPEECH_SERVICE_NAME", "")
         self.AZURE_SPEECH_SERVICE_REGION = os.getenv("AZURE_SPEECH_SERVICE_REGION")
+        self.SPEECH_RECOGNIZER_LANGUAGES = self.get_env_var_array(
```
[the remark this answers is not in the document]
superhindupur: I've already tested the default scenario - with the languages being [""] - and the behavior is to just fall back to detecting only english (which the CWYD default was before any of my changes). So for simplicity's sake, I think we should keep it like this. Let me know if you disagree.

infra/main.bicep
superhindupur: Thanks for pointing this out Ross, I've added this to container. 
I don't think language settings should be a part of the base settings, and rather should be per deployment. So since you said you're happy either way, I propose to keep the setting here rather than in web.bicep.

Makefile
```diff
@@ -42,9 +42,9 @@ build-frontend: ## 🏗️ Build the Frontend webapp
 	@echo -e "\e[34m$@\e[0m" || true
 	@cd code/frontend && npm install && npm run build
 
-unittest-frontend: ## 🏗️ Unit test the Frontend webapp
+unittest-frontend: build-frontend ## 🏗️ Unit test the Frontend webapp
```
ross-p-smith, line 45:
```suggestion
unittest-frontend: build-frontend ## 🧪 Unit test the Frontend webapp
```

😂😂😂

ross-p-smith (approved):

adamdougal (dismissed):
LGTM

ross-p-smith (dismissed):

cecheta (approved):
Looks great, thank you!

Issue #321, opened 2024-02-22
Add integrated vectorization as an option vs. push mechanism (including vectorizers) and keep, document both approaches


ross-p-smith:

gmndrg:
The accelerator today just supports to index using a function and push the data in the index using the SDK. All the chunking and everything else works by calling langchain libraries to take care of the chunking and such.
Today, we have support to do fixed-chunking and have integrated vectorization using pull indexers via: https://learn.microsoft.com/en-us/azure/search/vector-search-integrated-vectorization, so the request is to add an option for customers to use the current option (using Push API) or pull option (using built-in indexers to do the cracking, chunking, vectorize and index). This option removes the need to use AI doc intelligence for customers who have simple text or other files not supported by AI doc intelligence. 

There are a few examples of this implementation, here for the demo: https://github.com/Azure-Samples/azure-search-openai-demo/blob/main/scripts/prepdocs.py

Here for Python scripts as a sample: https://github.com/Azure/azure-search-vector-samples/tree/main/demo-python/code/integrated-vectorization

I hope this helps with the requirement definition.

prasbhargava1:
HI @gmndrg , wanted to confirm your thoughts on when to provide user an option to choose between push or pull (integrated vectorization)

- When the user is deploying the Solution accelerator, at the very beginning, via  .env file. So with this option, bicep will create a setup that will only work with only 1 of the 2 options.
- OR, User should be able to decide between push or pull while adding documents via admin portal. Here both the push and pull will coexist and user can make a decision at document level. (this solution will have higher complexity, we may have to check if data added in index via push api is not wiped off when the indexer runs.)

gmndrg:
Hi @prasbhargava1 it should be at the very beginning since the indexer is created for scheduled approach and with the fixed options and it would keep running. However, customer can also use push to the same index, both can coexist. So the current method of uploading data can remain there for uploading URLs and such but an indexer picking data from a container if needed. Please make sure to enable deletion policy by default: https://learn.microsoft.com/en-us/azure/search/search-howto-index-changed-deleted-blobs?tabs=portal.
Thank you so much

superhindupur:
[...]

PR #679, opened 2024-04-15
IV resources creation

komalg1:
## Purpose

* Required by [#321](https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/issues/321)
* Draft PR for resources creation via code

```
[ ] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...


infra/core/search/SetupSearchService.ps1
komalg1, line 1: Will remove this if we go ahead with resource creation via code

infra/core/search/search-service-iv.bicep
komalg1, line 1: Will remove this if we go ahead with resource creation via code

infra/main.bicep
```diff
@@ -320,6 +329,22 @@ module search './core/search/search-services.bicep' = {
   }
 }
 
+// module setupSearchService './core/search/search-service-iv.bicep' = if (azureSearchUseIntegratedVectorization) {
```
komalg1, line 332: This snippet will also be deleted

infra/core/search/SetupSearchService.ps1
```diff
@@ -0,0 +1,89 @@
...
+            };
+            'credentials' = @{
+                'connectionString' = $dataSourceConnectionString
+            };
+        }
+        $indexerDefinition = @{
+            'name' = $azureSearchIndexer;
+            'targetIndexName' = 'azureblob-index';
+            'dataSourceName' = 'azureblob-datasource';
+            'schedule' = @{ 'interval' = 'PT5M' };
```
adamdougal, line 48: Should we make the interval configurable? I can see that being useful.

adamdougal (commented):
I do think the code approach would be best. It allows us to test it and adapat it more easily. The downside is that this is called everytime and I'm not sure how much latency it adds, so maybe we can change it to run on startup, or check if we've already run it. We need to add test coverage to all of this too

code/backend/batch/utilities/helpers/AzureSearchIVDatasourceHelper.py
```diff
@@ -0,0 +1,43 @@
...
+                else DefaultAzureCredential()
+            ),
+        )
+        container = SearchIndexerDataContainer(
+            name=env_helper.AZURE_BLOB_CONTAINER_NAME
+        )
+        data_source_connection = SearchIndexerDataSourceConnection(
+            name=env_helper.AZURE_SEARCH_DATASOURCE_NAME,
+            type="azureblob",
+            connection_string=f"DefaultEndpointsProtocol=https;AccountName={env_helper.AZURE_BLOB_ACCOUNT_NAME};AccountKey={env_helper.AZURE_BLOB_ACCOUNT_KEY};EndpointSuffix=core.windows.net",
```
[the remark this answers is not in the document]
komalg1: Here datasource can be anything sql db, cosmos db etc. Could not find any docs in case the data source is using RBAC

Issue #101, opened 2023-12-11
Secure Speech Service Key/Region


hunterjam:
# Motivation

Ensure sensitive credentials cannot be exposed. We are currently returning the Azure Speech Key from the `/api/config` endpoint in plain text https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/ede0d6d93abd38c1b7b071ebf96ad18a3098b457/code/app.py#L32-L40

# How would you feel if this feature request was implemented?

![secure](https://media.giphy.com/media/v1.Y2lkPTc5MGI3NjExYzc3Y2QzMzBwYnVybWhuZG9yOWJibnNtY2Uza3Z0M24zZjN2NGJqOSZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/X68QCGb5qx596/giphy-downsized-large.gif)

# Requirements

- Stop returning the value of the Azure Speech Key from the `/api/config` endpoint. Consider using `SecretStr` from pydantic https://docs.pydantic.dev/2.6/examples/secrets/ to return a masked version of the secret
- Ensure that the speech service from the frontend is still functional (it calls the `/api/config` endpoint) #501 
- **Stretch:** Improve this endpoint to return the full config with all secrets masked

# Tasks

_To be filled in by the engineer picking up the issue_

- [ ] Implement https://github.com/Azure-Samples/cognitive-services-speech-sdk/blob/master/samples/js/browser/README.md#token-exchange-process
- [x] #683

adamdougal:
> The solution now has KeyVault - should this be utilised?

Looks like we are now using Key Vault. I think this is more that the key is returned in plain text from the `/api/config` end point 😬https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/ede0d6d93abd38c1b7b071ebf96ad18a3098b457/code/app.py#L37

I'll reword the issue to be clearer.

superhindupur:
https://github.com/Azure-Samples/cognitive-services-speech-sdk/blob/master/samples/js/browser/README.md#token-exchange-process can be relevant for this task

cecheta:
After discussing with the team, it has been decided to remove the `/api/config` endpoint entirely, as there is on requirement for the endpoint, and to have the endpoint while also masking secrets is non-trivial.

If such a requirement arises, the endpoint may be added back to the application.

PR #680, opened 2024-04-15
Add `/api/speech` endpoint

### LINKED ISSUE #501: Speech from UI does not work when using RBAC
# Describe the bug
The UI fetches the key for the Azure Speech Service from the `/api/config` endpoint, however there is no value for the key when using RBAC, therefore it fails. We need to somehow get speech service to work even when there is no key.

# Expected behavior
Can input query from the UI using voice.

# How does this bug make you feel?
😖

---

# Debugging information

## Steps to reproduce
* Deploy solution with `AZURE_AUTH_TYPE=rbac`
* Navigate to UI
* Press microphone
* Check console

## Screenshots

image

---

# Tasks

- [ ] Task 1
- [ ] Task 2
- [ ] ...

cecheta:
Closes #501, required by #101 

## Purpose

This PR adds a new `/api/speech` endpoint, which uses the Speech Key to return a short-lived token that the frontend can use to call the speech service. 
The frontend calls the endpoint every time the microphone button is pressed.
When using RBAC, the speech key is fetched directly from Azure as it is not present on any env var.

https://github.com/Azure-Samples/cognitive-services-speech-sdk/blob/master/samples/js/browser/README.md#token-exchange-process

The PR also refactors some of the tests/mocking in `test_app.py` ,so the mocking of the env_helper is done in one place.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
* Deploy application with Keys + with RBAC

## What to Check
* Test that speech-to-text is working


infra/main.bicep
```diff
@@ -398,6 +398,7 @@ module web './app/web.bicep' = if (hostingModel == 'code') {
       AZURE_OPENAI_API_VERSION: azureOpenAIApiVersion
       AZURE_OPENAI_STREAM: azureOpenAIStream
       AZURE_OPENAI_EMBEDDING_MODEL: azureOpenAIEmbeddingModel
+      AZURE_RESOURCE_GROUP: rgName
```
ross-p-smith, line 401: If we are going to make this change to every app service - we could move this down to `/infra/core/host/appservice.bicep`

ross-p-smith:
I think you've just made the same comment - but maybe some co-ordination with #660. It feels like maybe this goes in first and shows how to get config? What are your thoughts?

cecheta:
> I think you've just made the same comment - but maybe some co-ordination with #660. It feels like maybe this goes in first and shows how to get config? What are your thoughts?

Actually, this PR removes the ability to fetch config from the `/api/config` endpoint, but now I'm thinking it would actually be better to leave that in, so it can be used in #660 

Will update

superhindupur:
@cecheta so we're on the same page:
- I'm now convinced that injecting the env variable at buildtime into the frontend typescript app in https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/660 is overly complicating the build-time dependencies. 
- I agree that /api/config is a good place to add the speech languages 
- Should I go ahead and remove the azureSpeechKey from the /api/config endpoint in my PR so that we don't return the key from there?

infra/main.bicep
cecheta: Done

infra/main.bicep
```diff
@@ -877,8 +878,10 @@ output AZURE_SEARCH_TITLE_COLUMN string = azureSearchTitleColumn
 output AZURE_SEARCH_URL_COLUMN string = azureSearchUrlColumn
 output AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION bool = azureSearchUseIntegratedVectorization
 output AZURE_SEARCH_INDEX string = azureSearchIndex
+output AZURE_SPEECH_SERVICE_NAME string = speechServiceName
 output AZURE_SPEECH_SERVICE_REGION string = location
 output AZURE_SPEECH_SERVICE_KEY string = useKeyVault ? storekeys.outputs.SPEECH_KEY_NAME : ''
+output AZURE_SUBSCRIPTION_ID string = subscription().subscriptionId
```
ross-p-smith, line 884: is this already output by azd? Suppose doesnt hurt if we are explicit - I'll let you close this based on your opinion

code/create_app.py
```diff
@@ -285,6 +287,24 @@ def conversation_without_data(request, env_helper):
...
+def get_speech_key(env_helper: EnvHelper):
+    """
+    Get the Azure Speech key directly from Azure.
+    This is required to generate short-lived tokens when using RBAC.
+    """
+    client = CognitiveServicesManagementClient(
+        credential=DefaultAzureCredential(),
+        subscription_id=env_helper.AZURE_SUBSCRIPTION_ID,
+    )
+    keys = client.accounts.list_keys(
```
adamdougal, line 300: Does listing keys even work with the RBAC auth type?

code/create_app.py
```diff
@@ -285,6 +287,24 @@ def conversation_without_data(request, env_helper):
         )
 
 
+@functools.cache
```
ross-p-smith, line 290: Does this file get used when as a function? I'm going based on the dockerfile and I dont think it copies this in

code/frontend/src/pages/chat/Chat.tsx
```diff
@@ -155,41 +147,47 @@ const Chat = () => {
...
 
-  const startSpeechRecognition = (
-    subscriptionKey: string,
-    serviceRegion: string
-  ) => {
+  const fetchSpeechToken = async (): Promise<{ token: string, region: string; }> => {
+    try {
+      const response = await fetch("/api/speech/token");
+      if (!response.ok) {
+        throw new Error("Network response was not ok");
```
adamdougal, line 154: Should we log out the response to aid future debugging?

code/tests/functional/backend_api/tests/with_data/test_speech_token.py
```diff
@@ -0,0 +1,61 @@
...
+from tests.functional.backend_api.app_config import AppConfig
+from tests.functional.backend_api.request_matching import (
+    RequestMatcher,
+    verify_request_made,
+)
+
+pytestmark = pytest.mark.functional
+
+
+def test_speech_token_returned(app_url: str, app_config: AppConfig):
```
adamdougal, line 13: 👍

infra/main.bicep
cecheta: Good point, removed

code/create_app.py
cecheta: Would the function need this?

code/create_app.py
cecheta: Yes it does, we've already assigned the Cognitive Services User role so it has permissions

code/frontend/src/pages/chat/Chat.tsx
cecheta: Added

superhindupur:
@cecheta and I chatted offline - the conclusion was that we will rename the /api/config to /api/speech and return the token from there. Then I will add the languages to it under #660

adamdougal (approved):
noice

ross-p-smith (dismissed):
😎

code/frontend/src/pages/chat/Chat.tsx
```diff
@@ -134,62 +131,48 @@ const Chat = () => {
...
-        );
-      } else {
-        const recognizer = multiLingualSpeechRecognizer(subscriptionKey, serviceRegion, [
-          "en-US",
-          "fr-FR",
-          "de-DE",
-          "it-IT"
-        ]);
-        recognizerRef.current = recognizer; // Store the recognizer in the ref
+      const { token, region } = await fetchSpeechToken();
```
superhindupur, line 152: Hey does this mean that the config endpoint will be called every time the microphone is clicked?

superhindupur (approved):

ross-p-smith (approved):

adamdougal (approved):

Issue #508, opened 2024-03-21
App is taking lot of time to response as compare to last version


paragpatilxi:
# Describe the bug
App is taking lot of time to response as compare to last version which was taking average 3 second to respond but current one taking more than 20 sec to run LLM chain
# Expected behavior
response should be faster
# How does this bug make you feel?
_Share a gif from [giphy](https://giphy.com/) to tells us how you'd feel_

---

# Debugging information

## Logs

If applicable, add logs to help the engineer debug the problem.

---

# Tasks

_To be filled in by the engineer picking up the issue_

- [x] https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/682
- [x] #647
- [ ] #625
- [x] #601

PR #682, opened 2024-04-15
Make EnvHelper a singleton

adamdougal:
- The current setup means we create 23 instances of this when submitting a question via the UI
- A timed single question via the UI took 62 seconds to get an answer
- After this change, the same question took 8 seconds to get an answer
- Obviously this isn't very robust testing! But the improvement is obvious
- Decided to pick this up as it's having an affect on code design reviews

Required by https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/issues/508


cecheta (approved):
image

PR #691, opened 2024-04-15
Reintroduce docker-compose

adamdougal:
- This used to exist before we removed Docker in https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/377
- However, Docker has now been reintroduced by https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/505
- We've had feedback that it would be useful to also reintroduce Docker
  compose for local deployment for those that do not use VSCode and/or
dev containers


docker/docker-compose.yml
```diff
@@ -0,0 +1,21 @@
+version: "0.1"
+name: cwydsa
+services:
+  web:
+    image: fruoccopublic.azurecr.io/rag-webapp
+    ports:
+      - "8080:80"
+    env_file:
+      - ../.env
```
ross-p-smith, line 9: Is this going to work with the files that are generated by azd?

adamdougal: Not out of the box no. We could try and use the `AZURE_ENV_NAME` env var, but I don't think there are guarantees that will be set. Alternatively, we could have a script wrapper around this that calls `azd env list`, gets the default env, and passes that to docker compose e.g. `docker-compose --env-file .my-env  up`. WDYT?

ross-p-smith: Irrespective of whether people have Windows/Linux/DevContainers, then they are going to need a Search Service/OpenAi to communicate with. There is a line in the MakeFile (at the top) that will get you the correct .env file - maybe we should wrap the docker-compose in the MakeFile

adamdougal: Good point, I'll see what can be done with the Makefile 👍

Makefile
```diff
@@ -46,6 +46,9 @@ unittest-frontend: ## 🏗️ Unit test the Frontend webapp
 	@echo -e "\e[34m$@\e[0m" || true
 	@cd code/frontend && npm install && npm run build && npm run test
 
+docker-compose-up: ## 🐳 Run the docker-compose file
+	@cd docker && AZD_ENV_FILE=/workspaces/chat-with-your-data-solution-accelerator/.azure/cwydsa/.env docker-compose up
```
ross-p-smith, line 50:
```suggestion
	@cd docker && AZD_ENV_FILE=$(AZURE_ENV_FILE) docker-compose up
```

docs/LOCAL_DEPLOYMENT.md
```diff
@@ -101,6 +101,19 @@ You can deploy the full solution from local with the following command `azd depl
 |`azd deploy adminweb`     | A Streamlit app for the "admin" site where you can upload and explore your data.         |
 |`azd deploy function`     | A python function app processing requests.          |
 
+### Running All Services Locally Using Docker Compose
+
+To run all applications using Docker Compose, you first need a `.env` file containing the configuration for your
+provisioned resources. This file can be created manually at the root of the project. Alternatively, if resources were
+provisioned using `azd provision` or `azd up`, a `.env` file is automatically generated in the `.azure/<env-name>/.env`
```
ross-p-smith, line 108: Maybe talk about the `azd env list` command

infra/main.bicep
```diff
@@ -889,3 +889,4 @@ output AZURE_APP_SERVICE_HOSTING_MODEL string = hostingModel
 output FRONTEND_WEBSITE_NAME string = hostingModel == 'code' ? web.outputs.FRONTEND_API_URI : web_docker.outputs.FRONTEND_API_URI
 output ADMIN_WEBSITE_NAME string = hostingModel == 'code' ? adminweb.outputs.WEBSITE_ADMIN_URI : adminweb_docker.outputs.WEBSITE_ADMIN_URI
 output LOGLEVEL string = logLevel
+output AzureWebJobsStorage string = function.outputs.azureWebJobsStorage
```
ross-p-smith, line 892: Sorry - only just saw this - this will be expose the storage account key in logs - what do we use this for?

adamdougal: Ah it's a required configuration of the Backend App Functions. Without it set in the `.env` you get this error:
```
Missing value for AzureWebJobsStorage in local.settings.json. This is required for all triggers other than HTTP. You can run 'func azure functionapp fetch-app-settings ' or specify a connection string in local.settings.json
```

adamdougal: I guess it'll need to be manually added instead?

ross-p-smith: Can we add `func azure functionapp fetch-app-settings` to the makefile just after the azd env file bit - or maybe use the Emulator thingy that I cant remember it's name

cecheta: Can we add `func azure functionapp fetch-app-settings` as a postprovision hook?

docker/docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+  admin:
+    image: fruoccopublic.azurecr.io/rag-adminwebapp
+    ports:
+      - "8081:80"
+    env_file:
+      - path: ../.env
+        required: false
+      - path: $AZD_ENV_FILE
+        required: false
+  backend:
```
cecheta, line 21: I'm not sure how useful this is, because in order to invoke the endpoints it requires the `FUNCTION_KEY` env var. Unless you could write a script to fetch the key from Azure and then apply that to the environment?

cecheta: And then for the admin app we would need to configure `BACKEND_URL` to point to the function running locally, within Docker

infra/main.bicep
cecheta: Either way, we shouldn't be outputting secrets

ross-p-smith: I agree - step one get the docker-compose in with a guide how to get it working and then see if there is a way to make that easier if we get feedback

cecheta (dismissed):

ross-p-smith (approved):

docker/docker-compose.yml
adamdougal: i got it

docker/docker-compose.yml
```diff
@@ -0,0 +1,36 @@
...
+      - "8082:80"
+    environment:
+      - AzureWebJobsSecretStorageType=files
+    env_file:
+      - path: .env
+        required: false
+      - path: $AZD_ENV_FILE
+        required: false
+    volumes:
+      -  ${HOST_DOCKER_FOLDER:-.}/function-host.json:/azure-functions-host/Secrets/host.json
```
ross-p-smith, line 36: snazzy

adamdougal: This took me a while to figure out 😅 One of the downsides of dev containers!

docker/docker-compose.yml
cecheta: Fair play haha

image

PR #692, opened 2024-04-15
[Integrated Vectorization] PR to add url to blob storage

komalg1:
## Purpose

* Required by [#321](https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/issues/321)
* For URL embeddings using Integrated Vectorization, this PR:

1. checks if AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION is true
2. calls the url
3. Parses the HTML & CSS tags using BeautifulSoup
4. Uploads the parsed content to the blob storage.
5. Adds the url as 'source_url' field to the blob metadata which will be used as an index schema field to link the citations shown in search results.

```
[ ] Yes
[x] No
```

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
Enable AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION by setting it to 'true'
Add the url from the Admin page
```

## What to Check
* On checking the Azure Blob Storage, there should be a folder as 'https:'
* Verify the contents of the blob


infra/main.json
cecheta, line 1: I don't think this file should be changing?

code/backend/pages/01_Ingest_Data.py
```diff
@@ -75,21 +78,33 @@ def remote_convert_files_and_add_embeddings(process_all=False):
...
-            env_helper.BACKEND_URL, "/api/AddURLEmbeddings"
-        )
-        r = requests.post(url=backend_url, params=params, json=body)
-        if not r.ok:
-            raise ValueError(f"Error {r.status_code}: {r.text}")
-        else:
-            st.success(f"Embeddings added successfully for {url}")
+    if env_helper.AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION:
+        try:
+            urls = st.session_state["urls"].split("\n")
```
cecheta, line 83: This line is the same in both the if/else branches, could you extract it to above?

code/backend/pages/01_Ingest_Data.py
```diff
@@ -75,21 +78,33 @@ def remote_convert_files_and_add_embeddings(process_all=False):
...
+            for url in urls:
+                response = urllib.request.urlopen(url)
```
cecheta, line 85: Do we have to use `urllib.request`? Can we use `requests`?

code/backend/pages/01_Ingest_Data.py
```diff
@@ -75,21 +78,33 @@ def remote_convert_files_and_add_embeddings(process_all=False):
...
+                parsed_data = BeautifulSoup(response.read(), "html.parser")
+                with io.BytesIO(parsed_data.get_text().encode("utf-8")) as stream:
+                    upload_file(stream, url)
+                st.success(f"Embeddings added successfully for {url}")
```
cecheta, line 89: I think this log statement should be about a file being uploaded

komalg1: Updated it to be in sync with the description in UI (Add urls to the knowledge base)

code/backend/pages/01_Ingest_Data.py
komalg1: Updated & just read, `requests` is more efficient & handles errors codes better. Any other reason for using `requests`?

infra/main.json
komalg1: Got merge conflicts earlier. Updated it to match main.json in main

code/backend/pages/01_Ingest_Data.py
```diff
@@ -75,21 +77,32 @@ def remote_convert_files_and_add_embeddings(process_all=False):
...
-            raise ValueError(f"Error {r.status_code}: {r.text}")
-        else:
-            st.success(f"Embeddings added successfully for {url}")
+    if env_helper.AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION:
+        try:
+            for url in urls:
+                response = requests.get(url)
+                parsed_data = BeautifulSoup(response.content, "html.parser")
+                with io.BytesIO(parsed_data.get_text().encode("utf-8")) as stream:
+                    upload_file(stream, url)
```
adamdougal, line 87: Have you checked the the URL is stored correctly? I know there's some post processing that happens for it.

(though just noticed the metadata you're adding so probably all good)

code/backend/pages/01_Ingest_Data.py
```diff
@@ -75,21 +77,32 @@ def remote_convert_files_and_add_embeddings(process_all=False):
...
+                st.success(f"Url {url} added to knowledge base")
+        except Exception:
+            st.error(traceback.format_exc())
```
adamdougal, line 90: I know it's happening elsewhere, but we should avoid returning stack traces in the response. Could we instead log it and return a generic error response?

code/backend/pages/01_Ingest_Data.py
```diff
@@ -75,21 +77,32 @@ def remote_convert_files_and_add_embeddings(process_all=False):
...
-        backend_url = urllib.parse.urljoin(
-            env_helper.BACKEND_URL, "/api/AddURLEmbeddings"
-        )
-        r = requests.post(url=backend_url, params=params, json=body)
-        if not r.ok:
-            raise ValueError(f"Error {r.status_code}: {r.text}")
-        else:
-            st.success(f"Embeddings added successfully for {url}")
+    if env_helper.AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION:
+        try:
```
adamdougal, line 82: It might be worth pulling out the contents of the if/else into separate funtions, one for the download + upload approach and one for the existing add url embeddings approach.

adamdougal (changes requested):
Looking good so far! Can we add tests for this?

code/backend/pages/01_Ingest_Data.py
cecheta: It was mainly for consistency, we use `requests` everywhere else. Also it's a bit more intuitive than `urllib` (IMO)

code/backend/pages/01_Ingest_Data.py
```diff
@@ -75,21 +77,32 @@ def remote_convert_files_and_add_embeddings(process_all=False):
...
+            for url in urls:
+                response = requests.get(url)
```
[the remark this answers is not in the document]
komalg1: This is for converting the document to the langchain `Document` type. So might not be useful for this case.

code/backend/pages/01_Ingest_Data.py
komalg1: Yes verified, it is uploading the url data correctly. In case of push, it does all the chunking, processing once url is added. In case of pull, we only need it to get indexed. Let me know if I am missing anything

code/backend/pages/01_Ingest_Data.py
komalg1: Updated!

komalg1:
> Looking good so far! Can we add tests for this?

Will create another PR for this. None of the pages: Ingest, Explore & Delete have tests yet.

cecheta:
> > Looking good so far! Can we add tests for this?
> 
> Will create another PR for this. None of the pages: Ingest, Explore & Delete have tests yet.

I came across this: https://docs.streamlit.io/develop/concepts/app-testing/get-started

I think it should be its own epic

code/backend/pages/01_Ingest_Data.py
```diff
@@ -75,11 +79,31 @@ def remote_convert_files_and_add_embeddings(process_all=False):
...
+    try:
+        for url in urls:
+            response = requests.get(url)
+            parsed_data = BeautifulSoup(response.content, "html.parser")
+            with io.BytesIO(parsed_data.get_text().encode("utf-8")) as stream:
+                upload_file(stream, url)
+            st.success(f"Url {url} added to knowledge base")
+    except Exception:
+        logger.error(traceback.format_exc())
+        st.error("Exception occurred while adding URLs to the knowledge base.")
```
cecheta, line 99: What do you think?

```suggestion
    for url in urls:
        try:
            response = requests.get(url)
            parsed_data = BeautifulSoup(response.content, "html.parser")
            with io.BytesIO(parsed_data.get_text().encode("utf-8")) as stream:
                upload_file(stream, url)
            st.success(f"Url {url} added to knowledge base")
        except Exception:
            logger.error(traceback.format_exc())
            st.error(f"Exception occurred while adding {url} to the knowledge base.")
```

komalg1: Updated

cecheta (approved):
Nice :)

adamdougal (approved):
LGTM

ross-p-smith (approved):

PR #695, opened 2024-04-16
Resolve muiltithreading issues around EnvHelper singleton

adamdougal:
- We're occasionally seeing the below error when ingesting:
```
Exception while executing function: Functions.batch_push_results Result: Failure
Exception: AttributeError: 'EnvHelper' object has no attribute 'AZURE_BLOB_ACCOUNT_NAME'
Stack:   File "/azure-functions-host/workers/python/3.11/LINUX/X64/azure_functions_worker/dispatcher.py", line 505, in _handle__invocation_request
    call_result = await self._loop.run_in_executor(
has context menu
```
- I believe this is caused by multiple threads running at the same time
  - The first one creates the instance and starts populating the config
  - The second sees the instance the first one has created, but tries to use it before the config has been populated
- This change wraps the code in a lock to ensure only one thread at a time can get an instance of the EnvHelper

### DIFF
--- a/code/backend/batch/utilities/helpers/EnvHelper.py
+++ b/code/backend/batch/utilities/helpers/EnvHelper.py
@@ -1,5 +1,6 @@
 import os
 import logging
+import threading
 from dotenv import load_dotenv
 from azure.identity import DefaultAzureCredential, get_bearer_token_provider
 from azure.keyvault.secrets import SecretClient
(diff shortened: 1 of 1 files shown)


code/backend/batch/utilities/helpers/EnvHelper.py
```diff
@@ -9,12 +10,14 @@
...
     def __new__(cls):
-        if cls._instance is None:
-            cls._instance = super(EnvHelper, cls).__new__(cls)
-            cls._instance.__load_config()
-        return cls._instance
+        with cls._lock:
+            if cls._instance is None:
+                cls._instance = super(EnvHelper, cls).__new__(cls)
+                cls._instance.__load_config()
+            return cls._instance
```
cecheta, line 20: From what I've seen online, it seems to be more like this

```suggestion
        if cls._instance is None:
            with cls._lock:
                if cls._instance is None:
                    cls._instance = super(EnvHelper, cls).__new__(cls)
                    cls._instance.__load_config()
        return cls._instance
```

https://medium.com/analytics-vidhya/how-to-create-a-thread-safe-singleton-class-in-python-822e1170a7f6

adamdougal: The problem with that, is that the `cls._instance` for the second thread will **not** be none, so it will return it without entering the lock and waiting for the first thread to load the config.

cecheta: Ah, so we haven't solved anything 😅

cecheta (approved):
Nice!

PR #698, opened 2024-04-16
Change the tests github action to use devcontainers and make commands.

superhindupur:
## Purpose
This PR:
* Changes the tests github workflow to use devcontainers so we don't need to install poetry etc.
* Changes the workflow to use makefile commands instead of bespoke ones.
* Adds frontend unit tests to run on PR (frontend unit tests weren't part of this action before).

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x ] Refactoring (no functional changes, no api changes)
[ ] Other
```

## How to Test
Here's my test run from my feature branch: https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/actions/runs/8709085598/job/23887985675

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,6 +28,53 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Get coverage artifact ID
+        id: coverage-artifact
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const workflows = await github.rest.actions.listRepoWorkflows({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+            });
+
+            const workflowId = workflows.data.workflows.find(workflow => workflow.name === "${{ github.workflow }}")?.id;
+
+            if (!workflowId) return "";
+
+            const workflowRuns = await github.rest.actions.listWorkflowRuns({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              workflow_id: workflowId,
+              branch: "${{ github.base_ref }}",
+              event: "push",
+              status: "success",
+            });
+
+            return workflowRuns.data.workflow_runs[0]?.id ?? "";
+          result-encoding: string
+          retries: 3
+      - name: Download main coverage artifact
+        uses: actions/download-artifact@v4
+        if: steps.coverage-artifact.outputs.result != ''
+        continue-on-error: true # There is a chance that the artifact doesn't exist, or has expired
+        with:
+          name: coverage
+          path: "${{ github.workspace }}/coverage-main"
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          run-id: ${{ steps.coverage-artifact.outputs.result }}
+      - name: Get coverage from main
+        id: coverage-value
+        run: |
+          MIN_COVERAGE=0
+
+          if [[ -f "./coverage-main/coverage.xml" ]]; then
+            MIN_COVERAGE=$(grep -m 1 "<coverage" "./coverage-main/coverage.xml" | sed -E 's/.*line-rate="([^"]*)".*/\1/') # Extract the line rate from the XML
+            MIN_COVERAGE=$(awk "BEGIN {print int($MIN_COVERAGE * 100)}") # Turn into percentage, rounding down to avoid rounding issues
+          fi
+
+          echo "MIN_COVERAGE=$MIN_COVERAGE" >> "$GITHUB_OUTPUT"
+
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -77,15 +74,17 @@ jobs:
...
         with:
-          name: coverage
-          path: |
-            coverage-junit.xml
-            coverage.xml
-          if-no-files-found: error
-      - name: Run Functional tests
-        run: poetry run pytest -m "functional"
+          imageName: ghcr.io/azure-samples/chat-with-your-data-solution-accelerator
+          runCmd: make unittest optional_args="--junitxml=coverage-junit.xml --cov=. --cov-report xml:coverage.xml --cov-fail-under ${{ steps.coverage-value.outputs.MIN_COVERAGE }}"
```
superhindupur, line 82: Running unit tests separately from others so that we can continue to collect coverage and fail if lower than previous run.

.github/workflows/tests.yml
```diff
@@ -16,21 +16,18 @@ jobs:
...
-          cache-dependency-path: "**/pyproject.toml"
-      - name: Install Poetry
-        run: |
-          pip install poetry
-      - name: Install dependencies
-        run: |
-          poetry install
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
```
adamdougal, line 29: Are there any concerns about this running for forks/external contibutions? Or is this completely hidden from them?

.github/workflows/tests.yml
```diff
@@ -77,15 +74,17 @@ jobs:
...
-        run: poetry run pytest -m "functional"
+          imageName: ghcr.io/azure-samples/chat-with-your-data-solution-accelerator
+          runCmd: make unittest optional_args="--junitxml=coverage-junit.xml --cov=. --cov-report xml:coverage.xml --cov-fail-under ${{ steps.coverage-value.outputs.MIN_COVERAGE }}"
+          push: never
+
+      - name: Run other tests in dev container
+        uses: devcontainers/ci@v0.3
+        with:
+          imageName: ghcr.io/azure-samples/chat-with-your-data-solution-accelerator
+          runCmd: make ci
```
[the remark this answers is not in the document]
ross-p-smith: Completely get why you've done this as I assume the frontend tests will not run without cypress installed. The original reason for the unit tests not to be in the devcontainer were down to speed. They can run in seconds rather than minutes (downloading the docker image).

Happy both ways as consistency is great! :-)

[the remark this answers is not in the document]
adamdougal: I'm wondering if it's better to avoid using the `ci` task. If something fails now, you have to scan through a large amount of logs to see what went wrong. With each task separated into it's own step, you can easily see which part failed. You can also limit the logs to that failed task.

I'm wondering if it's worth removing/renaming the `ci` task. In the past I've used something like `check` to run all tests and linting locally, while having ci separated. Happy to have a call to discuss though!

.github/workflows/tests.yml
cecheta: I believe that forks don't have access to secrets at all, so this would fail

.github/workflows/tests.yml
```diff
@@ -77,15 +74,17 @@ jobs:
           fi
 
           echo "MIN_COVERAGE=$MIN_COVERAGE" >> "$GITHUB_OUTPUT"
-      - name: Run Unit tests
-        run: poetry run pytest -m "not azure and not functional" --junitxml=coverage-junit.xml --cov=. --cov-report xml:coverage.xml --cov-fail-under ${{ steps.coverage-value.outputs.MIN_COVERAGE }}
-      - uses: actions/upload-artifact@v4
```
cecheta, line 82: We've lost the publishing of the coverage artifact so we no longer have coverage commenting/fail thresholds. Is there any way we can get the coverage files out of the container so we can still publish the artifact?

.github/workflows/tests.yml
```diff
@@ -16,21 +16,18 @@ jobs:
   test_package:
     name: Tests
     runs-on: ubuntu-latest
+
     steps:
-      - uses: actions/checkout@v4
-      - name: Setup python
-        uses: actions/setup-python@v5
+      - name: Checkout code
```
cecheta, line 21: Do we still need to checkout?

superhindupur:
We had an internal discussion and concluded that we want to run the unit tests outside of devcontainers on PR, so we can validate that things run fine outside devcontainers. So I am canceling this PR and will create a new one which just adds the frontend unit tests to the existing PR workflow.

PR #704, opened 2024-04-17
Fix issue with env vars clashing with functional tests

adamdougal:
- Tests were failing locally for me due to a clash with azd env vars
- Change to set ALL env vars used
- This stops any predefined env vars from setting things we don't want
  in functional tests e.g. Key Vault
- Also ensure previous env vars are restored after tests

### DIFF
--- a/code/tests/functional/backend_api/app_config.py
+++ b/code/tests/functional/backend_api/app_config.py
@@ -5,7 +5,7 @@ logger = logging.getLogger(__name__)
 
 
 class AppConfig:
-    before_config: dict[str, str | None] = {}
+    before_config: dict[str, str] = {}
     config: dict[str, str | None] = {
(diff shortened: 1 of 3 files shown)


code/tests/functional/backend_api/app_config.py
```diff
@@ -5,23 +5,67 @@
 
 
 class AppConfig:
+    before_config: dict[str, str | None] = {}
```
cecheta, line 8:
```suggestion
    before_config: dict[str, str] = {}
```
I think

cecheta (approved):

code/tests/functional/backend_api/app_config.py
adamdougal: Good spot!

cecheta (approved):

PR #706, opened 2024-04-18
Update prompt for QuestionAnswerTool

cecheta:
## Purpose

* This PR updates the prompt used for the QuestionAnswerTool, to match the prompt used in Azure OpenAI On Your Data.

The prompt is broken into a number of system, user and assistant messages:
* `answering_system_prompt` - Main system prompt
* `examples`
  * `documents` - Example set of documents fetched
  * `user_question` - Example user question
  * `answer` - Example assistant answer
* Chat history (only when orchestrator is OpenAI, not LangChain)
* `answering_user_prompt` - User prompt containing retrieved documents and user question

The new parameters are configurable in the Admin app. There is JSON schema validation for the example documents.

image
image

There is also a checkbox, `use_on_your_data_format`, which determines whether the new, more sophisticated prompt from Azure OpenAI On Your Data will be used, or whether the old behaviour will be used. If the checkbox is unticked, the new parameters are disabled.

image

## Backwards compatibility

The code is designed to handle users who have previously saved their own config, who will now upgrade to the latest code.

* If the user has modified the prompt from the default, then:
  * `use_on_your_data_format` will be `False`, meaning the old LLM chain will be used
  * `answering_user_prompt` will be set to the modified prompt
* If the user has not modified the prompt:
  * `use_on_your_data_format` will be `True`
  * `answering_user_prompt` will be set to the (new) default, and the old value of `answering_prompt` will be ignored

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
* Run the code locally and in Azure

## What to Check
* Check that the new prompt is being used

image

* Check that the legacy prompt can be used if required

image


code/backend/batch/utilities/helpers/ConfigHelper.py
```diff
@@ -70,28 +81,64 @@ def __init__(self, logging: dict):
...
         if env_helper.LOAD_CONFIG_FROM_BLOB_STORAGE:
-            try:
-                blob_client = AzureBlobStorageClient(
-                    container_name=CONFIG_CONTAINER_NAME
-                )
-                config_file = blob_client.download_file("active.json")
-                config = Config(json.loads(config_file))
-            except Exception:
+            blob_client = AzureBlobStorageClient(container_name=CONFIG_CONTAINER_NAME)
+
```
ross-p-smith, line 123: This should speed things up quite a bit - many thanks (feel free to resolve once read!) :-)

code/backend/batch/utilities/helpers/ConfigHelper.py
```diff
@@ -105,6 +152,7 @@ def get_default_config():
 {chat_history}
```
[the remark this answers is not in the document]
cecheta: I 100% agree, just not sure where to put it. What do you think?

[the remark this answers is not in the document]
cecheta: I've put it in `./config/default.json` for now

superhindupur:
@cecheta and I had a discussion about how to maintain backwards compatibility with the old prompt format whilst also keeping the code readable, and here is the summary:

- The methods that generate the LLM chain will be called generate_llm_chain and generate_on_your_data_llm_chain rather than calling the old one legacy_generate_llm_chain.
- The config the tells us whether the user is using the new prompt format is now called  use_on_your_data_format rather than use_new_format.

The above are more descriptive than calling things "old" and "new". Will be useful for new developers maintaining the repo.

code/backend/batch/utilities/helpers/EnvHelper.py
```diff
@@ -175,6 +175,8 @@ def __load_config(self, **kwargs) -> None:
         self.ORCHESTRATION_STRATEGY = os.getenv(
             "ORCHESTRATION_STRATEGY", "openai_function"
         )
+        # Ensure this environment variable is set
+        os.environ["ORCHESTRATION_STRATEGY"] = self.ORCHESTRATION_STRATEGY
```
superhindupur, line 179: Is it worth writing a test to verify this? Is there any check elsewhere that will catch the deletion of this line 179, since otherwise the json file will not get this value?

superhindupur (changes requested):
Except for the request for a test case for env helper, LGTM.

adamdougal (approved):
LGTM!

code/backend/batch/utilities/helpers/EnvHelper.py
cecheta: Good idea

cecheta: So I reworked it to still use the env helper instead of relying on env vars directly, hopefully this is easier to maintain

superhindupur (approved):
Thanks Chinedum, LGTM.

ross-p-smith (approved):

ross-p-smith (approved):
Thanks for all your patience on this. Let's get this in and we can look at the intent_prompt as another issue

PR #709, opened 2024-04-18
IV resource creation

komalg1:
## Purpose

* Required by #321
* This PR adds helper classes for creating the Integrated Vectorization resources i.e index, indexer, skillset and datasource.
* It uses the existing `DocumentProcessor.process` method & checks if AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION is enabled

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
If AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION is True, when the user uploads a document, the search index, indexer, datasource & skillset should be created.
```

## What to Check
* Verify that the resources are created via the azure portal
* Verify that the document was indexed in the azure portal search resource


code/backend/batch/utilities/helpers/AzureSearchIVDatasourceHelper.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class AzureSearchIVDatasourceHelper:
+    def __init__(self, env_helper: EnvHelper):
+        self.env_helper = env_helper
+        self.indexer_client = SearchIndexerClient(
+            self.env_helper.AZURE_SEARCH_SERVICE,
+            (
+                AzureKeyCredential(self.env_helper.AZURE_SEARCH_KEY)
+                if self.env_helper.AZURE_AUTH_TYPE == "keys"
```
superhindupur, line 21: Can we extract `if self.env_helper.AZURE_AUTH_TYPE == "keys"` into a method, possibly in EnvHelper that is called something like, isAuthTypeKeys(), so that we don't leak the exact format of the "keys" string everywhere? Thanks!

superhindupur: In fact, it looks like we could also extract this logic of creating the SearchIndexerClient somewhere common for all the places using it.

code/backend/batch/utilities/helpers/AzureSearchIVSkillsetHelper.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def create_skillset(self):
+        skillset_name = f"{self.env_helper.AZURE_SEARCH_INDEX}-skillset"
+
+        split_skill = SplitSkill(
+            description="Split skill to chunk documents",
+            text_split_mode="pages",
+            context="/document",
+            maximum_page_length=2000,
+            page_overlap_length=500,
```
superhindupur, line 41: Can these sizes be made configurable? If we have an existing chunk size configuration, can we reuse them here?

code/backend/batch/utilities/helpers/AzureSearchIVIndexerHelper.py
```diff
@@ -0,0 +1,54 @@
...
+                ),
+                FieldMapping(
+                    source_field_name="metadata_storage_path",
+                    target_field_name="source",
+                ),
+            ],
+        )
+        self.indexer_client.create_or_update_indexer(indexer)
+        # Run the indexer
+        self.indexer_client.run_indexer(indexer_name)
```
adamdougal, line 44: Can this call the `run_indexer` function that you've created?

code/tests/utilities/helpers/test_AzureSearchIVIndexerHelper.py
```diff
@@ -0,0 +1,75 @@
...
+    env_helper_mock.AZURE_AUTH_TYPE = "rbac"
+    azure_search_indexer = AzureSearchIVIndexerHelper(env_helper_mock)
+
+    # when
+    azure_search_indexer.create_or_update_indexer("indexer_name", "skillset_name")
+
+    # then
+    azure_search_indexer.indexer_client.create_or_update_indexer.assert_called_once_with(
+        search_indexer_mock.return_value
+    )
```
adamdougal, line 75: Can we test the config being applied to the indexer?

adamdougal (changes requested):

code/tests/functional/backend_api/tests/with_data/test_conversation_custom.py
```diff
@@ -100,7 +100,11 @@ def test_post_makes_correct_call_to_get_search_index(
                 "Accept": "application/json;odata.metadata=minimal",
                 "Api-Key": app_config.get("AZURE_SEARCH_KEY"),
             },
-            query_string="api-version=2023-11-01",
+            query_string=(
+                "api-version=2023-11-01"
+                if not app_config.get("AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION")
+                else "api-version=2023-10-01-Preview"
```
superhindupur, line 106: Nit: can we change this to use "if app_config.get("AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION")" rather than "if not", will make it easier to read.

pyproject.toml
```diff
@@ -29,7 +29,7 @@ azure-storage-queue = "12.9.0"
 beautifulsoup4 = "4.12.3"
 fake-useragent = "1.5.1"
 chardet = "5.2.0"
-azure-search-documents = "11.4.0"
+azure-search-documents = "11.6.0b1"
```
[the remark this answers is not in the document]
komalg1: It is needed for Integrated Vectorisation

infra/main.bicep
```diff
@@ -781,6 +798,10 @@ module storage 'core/storage/storage-account.bicep' = {
       {
         name: blobContainerName
         publicAccess: 'None'
+        deleteRetentionPolicy: azureSearchUseIntegratedVectorization ? {
```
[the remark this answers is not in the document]
komalg1: Yes it is required for using indexers with azure storage as datasource. This is for delete detection (when blobs are deleted, the corresponding indexed documents are also removed). https://learn.microsoft.com/en-us/azure/search/search-howto-index-changed-deleted-blobs?tabs=portal
Note: All this information will be in ReadMe for Integrated Vectorisation

pyproject.toml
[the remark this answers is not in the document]
cecheta: Is it possible to use the latest beta version? It will probably be closest to the next stable version

code/backend/batch/utilities/helpers/EnvHelper.py
```diff
@@ -71,6 +71,10 @@ def __load_config(self, **kwargs) -> None:
             "AZURE_SEARCH_CONVERSATIONS_LOG_INDEX", "conversations"
         )
         # Integrated Vectorization
+        self.AZURE_SEARCH_DATASOURCE_NAME = os.getenv(
+            "AZURE_SEARCH_DATASOURCE_NAME", ""
+        )
+        self.AZURE_SEARCH_INDEXER_NAME = os.getenv("AZURE_SEARCH_INDEXER_NAME", "")
```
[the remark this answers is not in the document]
cecheta: We could/should probably make a test that catches this

code/backend/batch/utilities/helpers/DocumentProcessorHelper.py
```diff
@@ -22,20 +28,38 @@ def __init__(self):
         pass
 
     def process(self, source_url: str, processors: List[Processor]):
-        vector_store_helper = AzureSearchHelper()
-        vector_store = vector_store_helper.get_vector_store()
-        for processor in processors:
+        env_helper: EnvHelper = EnvHelper()
+        if env_helper.AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION:
```
cecheta, line 32: Could we have separate functions for whether IV is enabled or not?

code/backend/batch/utilities/helpers/DocumentProcessorHelper.py
```diff
@@ -22,20 +28,38 @@ def __init__(self):
...
-                document_chunking = DocumentChunking()
-                documents: List[SourceDocument] = []
-                documents = document_loading.load(source_url, processor.loading)
-                documents = document_chunking.chunk(documents, processor.chunking)
-                keys = list(map(lambda x: x.id, documents))
-                documents = [
-                    document.convert_to_langchain_document() for document in documents
-                ]
-                return vector_store.add_documents(documents=documents, keys=keys)
+                search_datasource_helper = AzureSearchIVDatasourceHelper(env_helper)
```
cecheta, line 34: I would suggest that we don't need to pass the env helper around now that it's a singleton, but happy to be overruled

code/backend/batch/utilities/helpers/DocumentProcessorHelper.py
```diff
@@ -22,20 +28,38 @@ def __init__(self):
         pass
 
     def process(self, source_url: str, processors: List[Processor]):
-        vector_store_helper = AzureSearchHelper()
-        vector_store = vector_store_helper.get_vector_store()
-        for processor in processors:
+        env_helper: EnvHelper = EnvHelper()
+        if env_helper.AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION:
             try:
-                document_loading = DocumentLoading()
```
cecheta, line 29: Am I right in saying that this code runs every time a new file is uploaded? Yet, the code doesn't use `source_url` at all

It seems to me like we would only want this code to run one time, once the infra has been provisioned, not every time a file is uploaded?

code/backend/batch/utilities/helpers/DocumentProcessorHelper.py
adamdougal: I think it's probably still worth it, it makes it easier to test, you can just pass the mock in rather than having the patch it in.

code/tests/utilities/helpers/test_AzureSearchIVSkillsetHelper.py
```diff
@@ -0,0 +1,135 @@
...
+        description="Skillset to chunk documents and generating embeddings",
+        skills=[
+            split_skill_mock.return_value,
+            azure_open_ai_embedding_skill_mock.return_value,
+        ],
+        index_projections=search_indexer_index_projections_mock.return_value,
+    )
+    search_indexer_client_mock.return_value.create_or_update_skillset.assert_called_once_with(
+        search_indexer_skillset_mock.return_value
+    )
```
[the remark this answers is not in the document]
komalg1: Do you mean we create a dummy skill set config and test against that? @adamdougal

code/backend/batch/utilities/helpers/DocumentProcessorHelper.py
cecheta: Or is it that we manually kick off the indexer every time a file is uploaded?

code/tests/utilities/helpers/test_AzureSearchIVSkillsetHelper.py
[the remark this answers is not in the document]
adamdougal: Ah I mean test that the `SplitSkill`,  `SearchIndexerIndexProjections` and `SearchIndexerSkillset` are being created correctly. At the moment they're mocked and we're only testing the creation of `AzureOpenAIEmbeddingSkill`.

infra/main.bicep
```diff
@@ -296,6 +302,17 @@ module searchServiceRoleOpenai 'core/security/role.bicep' = if (authType == 'rba
   }
 }
 
+// Storage Blob Data Reader
+module blobDataReaderRoleSearch 'core/security/role.bicep' = if (authType == 'rbac') {
```
cecheta, line 306: I think we need `Cognitive Services OpenAI User` as well, to generate embeddings

komalg1: It is already added https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/44bd0be460f498285daadf9ec2e9e0fc8b0f1dc5/infra/main.bicep#L835C1-L835C8

infra/main.bicep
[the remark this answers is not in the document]
komalg1: Updated to be used directly in `storage-account.bicep` . It is already defined as a top level param in the file

code/backend/batch/utilities/helpers/DocumentProcessorHelper.py
komalg1: This is the existing method for Push approach. Is called everytime a document is uploaded

code/backend/batch/utilities/helpers/DocumentProcessorHelper.py
komalg1: Updated

infra/main.bicep
```diff
@@ -885,6 +906,8 @@ output AZURE_SEARCH_TITLE_COLUMN string = azureSearchTitleColumn
 output AZURE_SEARCH_URL_COLUMN string = azureSearchUrlColumn
 output AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION bool = azureSearchUseIntegratedVectorization
 output AZURE_SEARCH_INDEX string = azureSearchIndex
+output AZURE_SEARCH_INDEXER_NAME string = azureSearchIndexer
```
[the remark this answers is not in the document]
komalg1: deliberately not setting it up

code/tests/functional/backend_api/tests/with_data/test_conversation_custom.py
```diff
@@ -100,7 +100,11 @@ def test_post_makes_correct_call_to_get_search_index(
                 "Accept": "application/json;odata.metadata=minimal",
                 "Api-Key": app_config.get("AZURE_SEARCH_KEY"),
             },
-            query_string="api-version=2023-11-01",
+            query_string=(
+                "api-version=2023-11-01"
+                if not app_config.get("AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION")
```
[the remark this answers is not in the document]
komalg1: Updated to `api-version=2023-10-01-Preview` because that will be the api version is the beta version of azure-search-documents is in main

code/backend/batch/utilities/helpers/EnvHelper.py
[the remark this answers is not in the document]
komalg1: Updated the env values for now! Will add a separate PR later for the whole app_config file test

code/backend/batch/utilities/helpers/DocumentProcessorHelper.py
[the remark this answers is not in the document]
komalg1: Added

code/backend/batch/utilities/helpers/AzureSearchIVIndexerHelper.py
```diff
@@ -0,0 +1,54 @@
...
+            self.env_helper.AZURE_SEARCH_SERVICE,
+            (
+                AzureKeyCredential(self.env_helper.AZURE_SEARCH_KEY)
+                if self.env_helper.AZURE_AUTH_TYPE == "keys"
+                else DefaultAzureCredential()
+            ),
+        )
+
+    def create_or_update_indexer(self, indexer_name: str, skillset_name: str):
+        indexer = SearchIndexer(
```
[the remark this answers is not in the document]
komalg1: No, we longer need to add a scheduler

code/backend/batch/utilities/helpers/AzureSearchIVIndexerHelper.py
komalg1: that function is still work in progress and is for 'Reprocess All' functionality

code/tests/utilities/helpers/test_AzureSearchIVDatasourceHelper.py
```diff
@@ -0,0 +1,89 @@
...
+    assert (
+        azure_search_iv_datasource_helper.indexer_client
+        == search_indexer_client_mock.return_value
+    )
+    search_indexer_data_container_mock.assert_called_once_with(
+        name=env_helper_mock.AZURE_BLOB_CONTAINER_NAME
+    )
+
+
+def test_create_or_update_datasource_rbac(
```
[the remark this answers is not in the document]
komalg1: Updated the tests

code/backend/batch/utilities/helpers/AzureSearchIVDatasourceHelper.py
komalg1: Updated it to call a function to get the Auth_type. SearchIndexerClient is being called only for IV, so I dont think it should be in EnvHelper class. Let me know your thoughts

code/backend/batch/BatchPushResults.py
```diff
@@ -53,5 +55,8 @@ def do_batch_push_results(msg: func.QueueMessage) -> None:
         )
     )
     # Process the file
-    document_processor.process(source_url=file_sas, processors=processors)
-    blob_client.upsert_blob_metadata(file_name, {"embeddings_added": "true"})
+    if env_helper.AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION:
+        document_processor.process_using_integrated_vectorisation(source_url=file_sas)
+    else:
+        document_processor.process(source_url=file_sas, processors=processors)
```
ross-p-smith, line 61: If this is now in this conditional block, should we pull in the piece of code that creates `processors` too - as that calls out to blob storage where it doesnt need to now

code/backend/batch/utilities/helpers/AzureSearchIVDatasourceHelper.py
```diff
@@ -0,0 +1,47 @@
+from azure.search.documents.indexes.models import (
```
ross-p-smith, line 1: How about a new folder in utilities called `integrated_vectorization` with these new classes and try and get rid of Helper and the long name?

ross-p-smith:
Fantastic Komal! 💖

infra/main.bicep
cecheta: That's the user role, we also need to give permissions to the Search Service

infra/main.bicep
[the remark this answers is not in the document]
cecheta: I think we generally keep the modules in the `core` directory generic and not specific to CWYD, so we shouldn't add any `azureSearchUseIntegratedVectorization`-related logic there

infra/main.bicep
```diff
@@ -154,6 +154,12 @@ param azureSearchSku string = 'standard'
 @description('Azure AI Search Index')
 param azureSearchIndex string = 'index-${resourceToken}'
 
+@description('Azure AI Search Indexer')
```
ross-p-smith, line 157: We have no way to turn this on in this PR - do we need to amend the allowable values for `azureSearchUseIntegratedVectorization` and then add something like 

`param azureSearchUseIntegratedVectorization = readEnvironmentVariable('AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION', false)`

into the main.bicepparam

ross-p-smith: Komal will raise another PR to do this

code/backend/batch/utilities/helpers/AzureSearchIVDatasourceHelper.py
komalg1: Agreed, Updated the names

infra/main.bicep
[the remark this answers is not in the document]
komalg1: Agreed, updated

code/backend/batch/utilities/integrated_vectorization/AzureSearchDatasource.py
```diff
@@ -0,0 +1,47 @@
...
+            connection_string=connection_string,
+            container=container,
+            data_deletion_detection_policy=NativeBlobSoftDeleteDeletionDetectionPolicy(),
+        )
+        self.indexer_client.create_or_update_data_source_connection(
+            data_source_connection
+        )
+
+    def generate_datasource_connection_string(self):
+        if self.env_helper.AZURE_AUTH_TYPE == "keys":
```
ross-p-smith, line 44: Could we use your new `self.env_helper.is_auth_type_keys()` method here too?

ross-p-smith (changes requested):

code/backend/batch/utilities/helpers/DocumentProcessorHelper.py
```diff
@@ -39,3 +46,22 @@ def process(self, source_url: str, processors: List[Processor]):
             except Exception as e:
                 logger.error(f"Error adding embeddings for {source_url}: {e}")
                 raise e
+
+    def process_using_integrated_vectorisation(self, source_url: str):
+        env_helper: EnvHelper = EnvHelper()
+        llm_helper: LLMHelper = LLMHelper()
+        try:
+            search_datasource_helper = AzureSearchDatasource(env_helper)
```
cecheta, line 54: Nit: These aren't helpers anymore 🙈

code/backend/batch/utilities/integrated_vectorization/AzureSearchIndex.py
```diff
@@ -0,0 +1,176 @@
...
+                    name="myOpenAI",
+                    kind="azureOpenAI",
+                    azure_open_ai_parameters=azure_open_ai_parameters,
+                ),
+            ],
+        )
+
+    def get_semantic_search_config(self):
+        semantic_config = SemanticConfiguration(
+            name="my-semantic-config",
```
cecheta, line 170: We could use `AZURE_SEARCH_SEMANTIC_SEARCH_CONFIG` from the env helper for the name

code/backend/batch/utilities/helpers/EnvHelper.py
```diff
@@ -71,9 +71,19 @@ def __load_config(self, **kwargs) -> None:
...
         self.AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION = self.get_env_var_bool(
             "AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION", "False"
         )
+        self.AZURE_SEARCH_IV_MAX_PAGE_LENGTH = os.getenv(
```
cecheta, line 81: Could you either make the defaults strings, or wrap the result in `int()`, to avoid typing issues in the future?

cecheta (commented):
Looking really good, a few small things from me

pyproject.toml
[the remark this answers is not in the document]
komalg1: I tested this (0b1) version is working well for both Push & Pull. Will raise another PR after testing again in case we need to update to the latest beta version. @cecheta thoughts? Can I resolve conversation?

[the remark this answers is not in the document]
cecheta: Yeah of course 😄

ross-p-smith (dismissed):
Thanks for persisting with this Komal! great work!

cecheta (dismissed):
🚀

cecheta (approved):

superhindupur:
Sorry I couldn't catch this before it was merged. Good stuff!

Issue #605, opened 2024-04-08
Upgrade Docker Build


ross-p-smith:
# Motivation

We currently build and push the 3 docker images on a commit to main. This means you could break the main build without knowing. Amending this will show Pull Request Authors problems earlier.

# How would you feel if this feature request was implemented?

![image](https://media.giphy.com/media/v1.Y2lkPTc5MGI3NjExeWxqZG1hcnh2OHM5dW1jdTA4Y2J6Z2ZweWphdjdpdWh0djBlZDJuMyZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/6AFldi5xJQYIo/giphy.gif)

# Requirements

A list of requirements to consider this feature delivered
- We should build the docker images when the Tests pass on all branches
- We should however only push when we are on main.
- We should move the current bash scripts to use this action https://github.com/docker/build-push-action
- We should parameterise the name of the Container Registry that we are publishing too.
- Add the following vscode extension: -  github.vscode-github-actions

# Tasks

_To be filled in by the engineer picking up the issue_

- [x] Replace existing scripts with official docker GitHub action. #[710](https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/710)
- [ ] Parameterize name of the registry [#712](https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/712) 
- [ ] Create GitHub actions to do docker build without push on PR creation. [#712](https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/712) 
- [ ] Add GitHub actions VSCode extension (This is already present in the [devcontainer configuration](https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/74352f234b29cbf1aace7d01c7adb311d7100e43/.devcontainer/devcontainer.json#L23))

PR #710, opened 2024-04-18
Update Docker build workflows to use the official action

superhindupur:
## Purpose
This PR replaces the custom script inside all three docker github workflows with the docker/build-push-action. This makes the github action configuration more readable and maintainable.

Required by #605.

```
[ ] Yes
[x ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x ] Other... Updates the CI/CD configuration.
```

## How to Test
Here's a passing build from my changes on my feature branch: https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/actions/runs/8741684840

### DIFF
--- a/.github/workflows/build-docker-admin.yml
+++ b/.github/workflows/build-docker-admin.yml
@@ -11,20 +11,31 @@ jobs:
     if: ${{ github.event.workflow_run.conclusion == 'success' }}
     runs-on: ubuntu-latest
     steps:
-   
+
+    - name: Checkout
+      uses: actions/checkout@v4
+
     - name: Docker Login
       uses: docker/login-action@v3
       with:
         registry: fruoccopublic.azurecr.io
         username: fruoccopublic
         password: ${{ secrets.DOCKER_PASSWORD }}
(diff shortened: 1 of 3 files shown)


ross-p-smith:
Great work! Is this just an initial replacement and there will be a further PR to change to only push when on main?
- build on all branches
- only push when on main

Maybe make it clear in the tasks section that this is just the part 1 if that is your intention

superhindupur:
That's correct. This just addresses part of the issue - just to replace the existing scripts with the official GitHub action. 

I will link this PR on the relevant task on the issue to make it clear, thanks.

adamdougal (approved):
LGTM!

I'm wondering if there's a way for us to reuse a single workflow and parameterise it, but we can look at that later.

superhindupur:
> LGTM!
> 
> I'm wondering if there's a way for us to reuse a single workflow and parameterise it, but we can look at that later.

Good idea, will keep that in mind as I make more change to this workflow under #605

PR #712, opened 2024-04-19
Create reusable github workflow for docker build and push to reduce duplication.

### LINKED ISSUE #605: Upgrade Docker Build

superhindupur:
## Purpose
We have three docker github workflows - one each for the frontend webapp, the backend (aka the frontend api) and the admin weapp.

* This PR extracts the common logic between these three workflows and parameterizes the registry, username and app names. 
* This PR also adds a workflow to run docker builds without push, that is configured to run on PR to main. This allows users to catch any issues with docker build before the code is merged to main.

Fixes #605. 

```
[ ] Yes
[x ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x ] Other... Updating CI/CD config for docker image creation.
```

## How to Test
Check the passing builds on this PR - they should contain the "build without push" workflow jobs.


.github/workflows/build-docker.yml
```diff
@@ -0,0 +1,56 @@
...
+        password: ${{ secrets.DOCKER_PASSWORD }}
+
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
+
+    - name: Get current date
+      id: date
+      run: echo "date=$(date +'%Y-%m-%d')" >> $GITHUB_OUTPUT
+
+    - name: Build and push the Docker image
```
adamdougal, line 47: Just wondering if we should rename this? Sometimes we don't push now, might be confusing

.github/workflows/build-docker-admin.yml
```diff
@@ -9,33 +9,11 @@ on:
...
-      uses: docker/build-push-action@v5
-      with:
-        context: .
-        file: docker/Frontend.Dockerfile
-        push: true
-        cache-from: type=registry,ref=fruoccopublic.azurecr.io/rag-adminwebapp:latest
-        tags: |
-          fruoccopublic.azurecr.io/rag-adminwebapp:latest
-          fruoccopublic.azurecr.io/rag-adminwebapp:${{ steps.date.outputs.date }}_${{ github.run_number }}
+    uses: ./.github/workflows/build-docker.yml
```
adamdougal, line 12: Should we merge these into a single workflow like you have with the "without push" workflow?

.github/workflows/build-docker.yml
superhindupur: Sure. How's "Reusable Docker build and optional push workflow"?

adamdougal: I think we can omit "Reusable", it gets outputted in the workflow run which would look a bit odd and we have it in the workflow name already.

Maybe "Build Docker Image (and push if on main branch)". Still not great but does what it says on the tin 🙈

adamdougal (approved):
LGTM!

.github/workflows/build-docker.yml
superhindupur: I'm calling it Build docker image and optionally push, since this file says nothing about main branch - that config is set by the workflow that uses it. That should satisfy your requirement of making it clear we don't always push - and it doesn't say "reusable". Sound good?

.github/workflows/build-without-push-docker.yml
```diff
@@ -0,0 +1,42 @@
+name: Docker Build without Push
+
+on:
+  pull_request:
+    branches: [main]
```
cecheta, line 5: I would have thought we don't want to run this one on `main`?

.github/workflows/build-without-push-docker.yml
```diff
@@ -0,0 +1,42 @@
...
+    types:
+    - opened
+    - ready_for_review
+    - reopened
+    - synchronize
+
+
+jobs:
+  docker-build-without-push-frontend:
+    uses: ./.github/workflows/build-docker.yml
```
cecheta, line 15: I think it would have been nice to use a matrix here

.github/workflows/build-without-push-docker.yml
superhindupur: Doesn't this mean - on "pull request to main"? And we wouldn't be creating a PR from main to main.

.github/workflows/build-without-push-docker.yml
superhindupur: I did consider matrices when I looked at the options. 
I didn't end up using them because we have combinations of configs - 3 dockerfiles each with their own tag labels. WIth a matrix, you end up with the cartesian product of the configs so it didn't quite fit in with our use case.

cecheta: What if you did something like

```yaml
matrix:
  build:
    - name: rag-adminwebapp
      dockerfile: docker/Admin.Dockerfile
    - name: rag-backend
      dockerfile: docker/Backend.Dockerfile
    - name: rag-webapp
      dockerfile: docker/Frontend.Dockerfile
```

That would only create 3 combinations?

Issue #323, opened 2024-02-22
Include GPT-4 V model to be able to search for images and embedding images.


ross-p-smith:
# Motivation

Company data often comprises various types of images, including screenshots, maps, and diagrams. By enabling the chat admin app to ingest and process these images, it can provide more accurate and relevant responses to user queries that involve visual data. This ensures that the chat app can fully utilise all available company data to deliver an improved user experience.

Note: Image processing is only available using GPT-4 https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/models#gpt-4-and-gpt-4-turbo-preview

# How would you feel if this feature request was implemented?

![gif](https://media.giphy.com/media/v1.Y2lkPTc5MGI3NjExeTdtZ3ZmZ29uNWV4dW5zaXZsNjI4ajBkcmRjam1nbmpuMmFteGp0ciZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/HHWdyVKaKJfI8qqJKU/giphy.gif)

# Requirements

- Ensure existing application works correctly with GPT-4
- Allow images to be uploaded via the Admin application
- When "Reprocess all" is click via the Admin app, reprocess the images
- When a question is asked, image data should be searched and passed to `gpt-4-vision` to generate a response
- Citations should link to the image stored in blob storage
- ~**Stretch:** Fallback to OCR/document intelligence if image of a document detected~
- ~**Stretch:** Allow images to be uploaded when chatting~

# Tasks

- [x] #713
- [x] #715
- [x] #728
- [x] #748
- [x] #749
- [x] #750
- [x] #752
- [x] #965
- [ ] #964
- [ ] #993
- [ ] Investigate it's possible to turn advanced image processing on/off without reprovisioning
- [ ] Allow images to be uploaded via pull model (if possible, this indicates it may not be - https://github.com/Azure-Samples/azure-search-openai-demo/blob/main/docs/gpt4v.md#setup-and-usage)

# Bugs

- [x] #929

ross-p-smith:
Reference here: - https://github.com/Azure-Samples/azure-search-openai-demo/pull/1056

adamdougal:
**Update 23rd April:**

- The computer vision and gpt-4-vision model deployment resources are now being provisioned
- This is applied if `USE_GPT4_VISION=true`
- Unfortunately, `gpt-4-vision` does not support function calling, so this is an additional deployment alongside another model
[...]

PR #714, opened 2024-04-19
Add spike docs for using computer vision

### LINKED ISSUE #713: Spike possible image data usage implementations
Required by https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/issues/323

# Description

Spike how to implement the usage of image data

# Tasks

- [ ] Generate doc detailing how to injest and chat with image data - https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/714

adamdougal:
- Includes a working Jupyter notebook demonstrating how to
ingest and chat with image based data

Closes #713


docs/spikes/using-image-data/ai-vision.ipynb
```diff
@@ -0,0 +1,530 @@
+{
```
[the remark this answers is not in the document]
superhindupur: Yep, that's what I'm doing. Reviewing this here is impossible :D

docs/spikes/using-image-data/ai-vision.ipynb
```diff
@@ -0,0 +1,530 @@
...
+    "4. Generate a caption of the image using `gpt-4-vision`\n",
+    "5. Generate embeddings of the caption using `text-embeddings-002-ada`\n",
+    "6. Store data in search index\n",
+    "\n",
+    "\n",
+    "### Question\n",
+    "\n",
+    "To ask a question using this data, the following steps are performed:\n",
+    "1. Generate embeddings for the question using `text-embeddings-002-ada`\n",
+    "2. Generate image imbeddings for the question using computer vision\n",
```
superhindupur, line 60: This second point seems to suggest that the question can be in the form of an image too, not just the data - is this correct? If yes, I'd suggest making this clear in the introduction sentence.

superhindupur: Ok, I think we need a little more explanation here. Why do we need to generate embeddings both using the ada model and computer vision?

pyproject.toml
```diff
@@ -47,6 +47,7 @@ black = "24.4.0"
 pre-commit = "3.7.0"
 pytest_httpserver = "1.0.10"
 trustme = "1.1.0"
+jupyter = "1.0.0"
```
cecheta, line 50: Should we add `ipykernel` as well?

docs/spikes/using-image-data/ai-vision.ipynb
superhindupur: What I would've thought is, since we're finally creating embeddings for the ingested image using the ada model (after the internediate steps of captioning etc), it should suffice to create the ada embeddings for the text question in order to make searching possible. Let me know what I'm missing here.

adamdougal: Ah so the `vectorizeImage` computer vision API uses a different embedding model then `text-embeddings-ada-002`. So to search for that image using a text based question, you need to use the `vectorizeText` computer vision API. You can't use `text-embeddings-ada-002` to embed the question and search for images as the vectors are constructed differently, including the size.

Using `gpt-4-vision` to generate a description and `text-embeddings-ada-002` to embed that description is not required, but provides extra context and better results, which I found especially useful for the flow chart.

I'll add a section explaining this now.

adamdougal:
> I'd like to understand what changes we will be making (and where) in Chat With Your Data based on what we learned from this spike. Doesn't have to be part of this PR, as long as it is written somewhere, even if just a github issue.

I've not looked closely at the code as yet. It won't look exactly like the code here as in reality we have things like openai functions and langchain to factor in. However, I've written up a task list here https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/issues/323, I'll then look more in detail at what's required for each step when I get to that point. Does that suffice?

docs/spikes/using-image-data/ai-vision.ipynb
```diff
@@ -56,16 +48,27 @@
...
+    "\n",
+    "## Why do we need two different embedding models?\n",
+    "\n",
+    "It is not required to use two different embedding models, however, using both Azure Computer Vision to embed the image\n",
+    "and `gpt-4-vision` to generate a description that is then embedded by `text-embeddings-002-ada` provides richer data and\n",
+    "provides better search results. This is useful in particular for diagrams and flow charts which show relationships and\n",
+    "decision points.\n",
+    "\n",
+    "When embedding the question, we need to embed it with both embeddings models to allow us to search over the previously\n",
+    "generated vectors. This is due to the embeddings models generating differently sized and formatted vectors."
```
superhindupur, line 66: Ok, so we're embedding both the image and the question with both models now.

In the implementation in CWYD, do we plan to embed all incoming questions with both models? Because when embedding the question we will not necessarily know if the answer is in an image or in a word doc for example.

superhindupur: Yes that makes sense. Thanks Adam. I think that answers all my questions.

superhindupur (approved):

cecheta (approved):
I enjoyed following this through, thanks!

ross-p-smith (approved):
loveit

docs/spikes/using-image-data/ai-vision.ipynb
```diff
@@ -0,0 +1,533 @@
...
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "e0da0335-4787-499a-abe9-bd99971bb98c",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Common imports\n",
+    "import re\n",
+    "import datetime\n",
+    "import base64\n",
```
ross-p-smith, line 79: nit: you dont use base64 - happy for you to ignore so you dont have to get reapprovals! :-)

PR #716, opened 2024-04-22
Make docker login conditional on push for docker github actions.

superhindupur:
## Purpose
Changing the build-without-push github action to only run the docker login step if we need to push. 
Rationale: We don't need to do this if we aren't pushing, plus forcing docker login on every PR breaks the builds for forks which don't have the docker password secret. 

```
[ ] Yes
[ x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ x] Other... CI/CD config update
```

## How to Test
Verify the docker builds on this PR.

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -20,7 +20,7 @@ on:
         type: boolean
     secrets:
       DOCKER_PASSWORD:
-        required: true
+        required: false
 
 jobs:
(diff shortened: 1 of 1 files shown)


cecheta (approved):
😎

adamdougal (approved):

ross-p-smith (approved):

Issue #635, opened 2024-04-09
Configurable Embedding capacity


ross-p-smith:
# Motivation

When we ingest multiple files, we often hit a quota issue. Allow the embedding capacity to be configurable.

# How would you feel if this feature request was implemented?

[!image](https://media.giphy.com/media/v1.Y2lkPTc5MGI3NjExbHAyMDMwNjEzMmtxbGwxdzZyN3NxdXAwYjRjaHZuem1ncHh3MjdyeiZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/E5auPRQpuG2vs1c7lN/giphy.gif) to tells us how you'd feel_

# Requirements

A list of requirements to consider this feature delivered
- Surface the embedding capacity as a parameter

# Tasks

_To be filled in by the engineer picking up the issue_

- [ ] Task 1
- [ ] Task 2
- [ ] ...

Issue #715, opened 2024-04-22
Provision AI services for image data


adamdougal:
Required by https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/issues/323

# Description

Provision the Azure services required to process image data

# Tasks

- [x] Provision Computer Vision Resource if gpt-4-vision being used - https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/722
- [x] Add required roles for RBAC auth type deployments - All apps already have the `Cognitive Services User`/`a97b65f3-24c7-4388-baec-2e87135dc908` role

adamdougal:
Looks like we already have the required roles to interact with the new computer vision resource.

PR #722, opened 2024-04-22
Provision Computer Vision if using GPT-4-vision

adamdougal:
- Creates required resources if using GPT-4-vision
- Allows capacity to be set, the quota for GPT-4-vision is 30, so our default capacity used all of it in one deployment

Required by https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/issues/715
Fixes https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/issues/635

OpenAI:
image

Computer Vision:
image


infra/main.bicep
```diff
@@ -138,6 +141,52 @@ param azureOpenAIEmbeddingModel string = 'text-embedding-ada-002'
...
+
+@description('Name of Computer Vision Resource SKU (if using GPT-4 and Vision Preview)')
+@allowed([
+  'F0'
+  'S1'
+])
+param computerVisionSkuName string = 'S1'
+
+@description('Location of Computer Vision Resource (if using GPT-4 and Vision Preview)')
+@allowed([// List taken from https://azure.microsoft.com/en-us/pricing/details/cognitive-services/computer-vision/
```
cecheta, line 158: I got an error that my region wasn't supported when I used UK South

I think we should use these regions: https://learn.microsoft.com/en-us/azure/ai-services/computer-vision/how-to/image-retrieval?tabs=csharp#:~:text=Be%20sure%20to%20create%20it%20in%20one%20of%20the%20permitted%20geographic%20regions%3A%20East%20US%2C%20France%20Central%2C%20Korea%20Central%2C%20North%20Europe%2C%20Southeast%20Asia%2C%20West%20Europe%2C%20West%20US.

infra/main.bicep
```diff
@@ -218,6 +267,10 @@ var tags = { 'azd-env-name': environmentName }
 var rgName = 'rg-${environmentName}'
 var keyVaultName = 'kv-${resourceToken}'
 
+var isGPT4 = azureOpenAIModel == 'gpt-4'
+var isVision = azureOpenAIModelVersion == 'vision-preview'
+var useGPT4Vision = isGPT4 && isVision
```
ross-p-smith, line 272: Is this going to be called Turbo going forward?

infra/main.bicep
```diff
@@ -258,7 +311,7 @@ module openai 'core/ai/cognitiveservices.bicep' = {
         }
         sku: {
           name: 'Standard'
-          capacity: 30
+          capacity: azureOpenAIModelCapacity
```
ross-p-smith, line 314: Linking #635

infra/main.bicep
```diff
@@ -218,6 +267,10 @@ var tags = { 'azd-env-name': environmentName }
 var rgName = 'rg-${environmentName}'
 var keyVaultName = 'kv-${resourceToken}'
 
+var isGPT4 = azureOpenAIModel == 'gpt-4'
+var isVision = azureOpenAIModelVersion == 'vision-preview'
```
superhindupur, line 271: Can isVision ever be true when isGPT4 is false?

Also, I'm thinking, to keep things simple, can we not force azureOpenAIModelVersion to be vision-preview when using GPT4? Is there a reason a user would definitely not want to use vision-preview when using GPT4? 

This way, we can just check if GPT4 is being used and assume vision is included.

infra/main.bicep
```diff
@@ -104,6 +104,9 @@ param azureOpenAIModelName string = 'gpt-35-turbo'
 
 param azureOpenAIModelVersion string = '0613'
 
+@description('Azure OpenAI Model Capacity')
+param azureOpenAIModelCapacity int = 30
```
superhindupur, line 108: Can we add a one liner explaining what this is?

infra/main.bicep
adamdougal: I'll add it to the param file so we can close this off while we're here.

infra/main.bicep
adamdougal: > Can isVision ever be true when isGPT4 is false?

Not currently. But this may change in the future.

> Also, I'm thinking, to keep things simple, can we not force azureOpenAIModelVersion to be vision-preview when using GPT4? Is there a reason a user would definitely not want to use vision-preview when using GPT4?

I don't think we want to force vision if gpt-4 is being used, users may want to use different models or may be restricted based on region. https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/models#model-summary-table-and-region-availability

infra/main.bicep
adamdougal: One liner where? A code comment or in the description? Feels like it may be better suited to a configuration page/doc.

Maybe we can just link to this page https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/quota ? WDYT?

infra/main.bicep
```diff
@@ -138,6 +141,31 @@ param azureOpenAIEmbeddingModel string = 'text-embedding-ada-002'
...
+@allowed([// List taken from https://learn.microsoft.com/en-us/azure/ai-services/computer-vision/how-to/image-retrieval?tabs=python#prerequisites
+  'eastus'
+  'westus'
+  'koreacentral'
+  'francecentral'
+  'northeurope'
+  'westeurope'
+  'southeastasia'
+])
+param computerVisionLocation string = location
```
cecheta, line 167: If someone has previously deployed the application to a region not in this list, the next time they deploy they'll get an error, even if they're not using vision. Is there any way we can avoid this?

adamdougal: Hmm, very very good catch! I guess one option is to not set a default, which would mean the user would always have to supply this. But that's not a great user exeperience for when they're not even going to be use Computer Vision. 

Maybe we should just skip validation of this for now? I'm not a huge fan of comments hidden in "code", but maybe that could be a best effort for now?

adamdougal: Alternatively, I could create a new `computervision.bicep`, that calls `cognitiveservices.bicep` and validate the param in there. That way it'll only be called if it's enabled. 

I _think_ that will work.

WDYT?

adamdougal: I get this error when I try and do that:

This symbol cannot be referenced here. Only other parameters can be referenced in parameter default values.bicep(BCP072)

adamdougal: Ah but then we get this error:
Expected a value of type "'eastus' | 'francecentral' | 'koreacentral' | 'northeurope' | 'southeastasia' | 'westeurope' | 'westus'" but the provided value is of type "'' | string".bicep(BCP033)

cecheta: Add `''` to the allowed values

adamdougal: > Add '' to the allowed values

I'm not sure I like that 😅  Lemme try the approach of moving the validation down

adamdougal: Yea that doesn't work, lets go with adding `''` to the allowed values

adamdougal:
I discovered the `gpt-4-vision` does not support function calling. Therefore changing this PR to **add** a `gpt-4-vision` deployment in addition to another model instead of replacing. https://platform.openai.com/docs/guides/function-calling/supported-models

adamdougal:
Screenshot from deployment using the latest code:

image

cecheta (approved):
Nice work

Issue #448, opened 2024-03-18
Once we delete file from the Admin Site "Delete Data", the file deleted from the vector DB, but its not deleted from the blob storage automatically.


sreedharguda:
# Describe the bug

# Expected behavior

# How does this bug make you feel?
_Share a gif from [giphy](https://giphy.com/) to tells us how you'd feel_

---

# Debugging information

## Logs

If applicable, add logs to help the engineer debug the problem.

---

# Tasks

_To be filled in by the engineer picking up the issue_

- [ ] Task 1
- [ ] Task 2
- [ ] ...

cecheta:
Hello @sreedharguda , thanks for raising this issue, we will look into it.

Please in the future, could you add more information to the issue description using the template rather than just the title, as this helps when reviewing the issue, thank you.

adamdougal:
We should consider moving the deletion to the backend function so that files are still removed from the index when deleting from blob storage directly.

PR #733, opened 2024-04-22
Delete files from blob storage

adamdougal:
Fixes https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/issues/448

### DIFF
--- a/code/backend/pages/03_Delete_Data.py
+++ b/code/backend/pages/03_Delete_Data.py
@@ -3,6 +3,7 @@ import os
 import traceback
 import sys
 from batch.utilities.helpers.AzureSearchHelper import AzureSearchHelper
+from batch.utilities.helpers.AzureBlobStorageHelper import AzureBlobStorageClient
 from dotenv import load_dotenv
 
 sys.path.append(os.path.join(os.path.dirname(__file__), ".."))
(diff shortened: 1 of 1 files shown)


code/backend/pages/03_Delete_Data.py
```diff
@@ -78,7 +80,23 @@ def delete_files(files):
     st.success("Deleted files: " + str(files_to_delete))
 
 
+def get_filename(file: str) -> str:
```
ross-p-smith, line 83: Just parsing the code in my head - will this work with @komalg1 IV work where the blob could be stored in a URI format?

adamdougal: Ah good point. Let me check on that!

code/backend/pages/03_Delete_Data.py
```diff
@@ -78,7 +80,23 @@ def delete_files(files):
...
+    /container_name/file_name
+    to return the file name.
+
+    Args:
+        file (str): The file path of the file to delete.
+
+    Returns:
+        filename (str): The filename part of the file path
+    """
+    return file.split("/")[2]
```
superhindupur, line 95: What is the behavior if, for some reason, the split or the [2] doesn't work? Will the code throw an error and never get to the search_client.delete_documents()?

adamdougal:
I think I'm going to hold off on this for now. Quite a few moving parts and more involved then anticipated for a quick win!

PR #739, opened 2024-04-23
Add Reset configuration button to Admin app

### LINKED ISSUE #738: Add "Reset to defaults" button for Admin app
# Motivation

At present, there is no easy way to revert config to the default once it has been changed in the Admin app. There should be a button to reset back to defaults.

# Requirements

- Add button to Admin app to reset config

# Tasks

cecheta:
Closes #738 

## Purpose

* Adds `Reset confiiguration to defaults` button so the configuration can be reset to the default.

It was observed when implementing #706 that there was not an easy way to reset the configuration once it had been modified, other than going into the source code.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
* Run admin app locally

https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/assets/60179183/d7faeb3e-d2a4-4fe4-8c61-d1622888c5a2

### DIFF
--- a/code/backend/pages/04_Configuration.py
+++ b/code/backend/pages/04_Configuration.py
@@ -340,5 +340,16 @@ try:
         ConfigHelper.save_config_as_active(current_config)
         st.success("Configuration saved successfully!")
 
+    with st.popover(":red[Reset confiiguration to defaults]"):
+        st.write(
+            "**Resetting the configuration cannot be reversed, proceed with caution!**"
+        )
+        name = st.text_input('Enter "reset" to proceed', key="reset_configuration")
+        if st.button(
+            ":red[Reset]", disabled=st.session_state["reset_configuration"] != "reset"
+        ):
+            ConfigHelper.delete_config()
+            st.success("Configuration reset successfully! Refresh to update.")
+
 except Exception:
     st.error(traceback.format_exc())
--- a/code/backend/batch/utilities/helpers/ConfigHelper.py
+++ b/code/backend/batch/utilities/helpers/ConfigHelper.py
@@ -163,3 +163,8 @@ class ConfigHelper:
                 )
 
         return ConfigHelper._default_config
+
+    @staticmethod
+    def delete_config():
+        blob_client = AzureBlobStorageClient(container_name=CONFIG_CONTAINER_NAME)
+        blob_client.delete_file(CONFIG_FILE_NAME)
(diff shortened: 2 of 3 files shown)


code/backend/batch/utilities/helpers/AzureBlobStorageHelper.py
```diff
@@ -204,14 +203,9 @@ def get_all_files(self):
         return files
 
     def upsert_blob_metadata(self, file_name, metadata):
-        if self.auth_type == "rbac":
```
cecheta, line 207: Not strictly related to this PR, but we already have a client created in the constructor, therefore we can use it here without needing if/else logic

ross-p-smith (dismissed):

code/backend/pages/04_Configuration.py
```diff
@@ -340,5 +341,21 @@ def validate_documents():
...
+            "**Resetting the configuration cannot be reversed, proceed with caution!**"
+        )
+        name = st.text_input('Enter "reset" to proceed', key="reset_configuration")
+        if st.button(
+            ":red[Reset]", disabled=st.session_state["reset_configuration"] != "reset"
+        ):
+            try:
+                ConfigHelper.delete_config()
+            except ResourceNotFoundError:
+                pass
```
superhindupur, line 355: We might want to surface this error up to the user. If resetting the configuration failed

superhindupur: I see now that this is only when the resource in the lob storage isn't found. Still it might make sense to let the user know something unexpected happened

code/backend/pages/04_Configuration.py
```diff
@@ -340,5 +341,21 @@ def validate_documents():
...
+
+            st.success("Configuration reset successfully! Refresh to update.")
```
superhindupur, line 357: Can we instead programmatically force a reload of the page?

code/backend/pages/04_Configuration.py
superhindupur: Yes, that sounds good!

code/backend/pages/04_Configuration.py
cecheta: Managed to find a way!

superhindupur (approved):

PR #740, opened 2024-04-23
Create an ADR for deciding whether or not to add PromptFlow.

superhindupur:
## Purpose
Creates an ADR for deciding whether to include promptflow to this repo. 

```
[ ] Yes
[x ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x ] Documentation content changes
[ ] Other... Please describe:
```

## How to Test
N/A

### DIFF
--- a/docs/design/adrs/2024-03-18-use-cypres-as-a-framework-for-e2e-testing.md
+++ b/docs/design/adrs/2024-03-18-use-cypres-as-a-framework-for-e2e-testing.md
@@ -1,6 +1,6 @@
 # Use Cypress as a framework for E2E testing
 
-* **Status:** proposed
+* **Status:** approved
 * **Proposer:** @superhindupur
 * **Date:** 2024-03-18
 * **Technical Story:** [Add E2E test suite](https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/issues/419)
@@ -8,10 +8,10 @@
 ## Context and Problem Statement
 
 E2E tests help test the deployed system as a whole, usually from the browser, for web applications like the Chat With Your Data website and admin website.
-They provide confidence in the overall functionality of the system, including how the various systems interact with each other. 
-Since these tests can be slow and complex to set up, we should ideally run a small subset of test cases as E2E tests. 
+They provide confidence in the overall functionality of the system, including how the various systems interact with each other.
+Since these tests can be slow and complex to set up, we should ideally run a small subset of test cases as E2E tests.
 
-There are several E2E UI testing frameworks out there, and this document looks at which one to use. 
+There are several E2E UI testing frameworks out there, and this document looks at which one to use.
 
 ## Decision Drivers
 
@@ -25,15 +25,15 @@ There are several E2E UI testing frameworks out there, and this document looks a
 * Selenium
 * Cypress
 * TestCafe
-* Several other options not mentioned here - like TestComplete, Robot Framework etc. They are not lightweight as they support mobile app testing in addition to browser testing, which we do not need. 
+* Several other options not mentioned here - like TestComplete, Robot Framework etc. They are not lightweight as they support mobile app testing in addition to browser testing, which we do not need.
 
 ## Decision Outcome
 
-Cypress, because it meets all the requirements. 
+Cypress, because it meets all the requirements.
 
 ## Pros and Cons of the Options
 
-### Selenium 
+### Selenium
 
 [Selenium](https://www.selenium.dev/) is one of the most widely used UI testing frameworks. It uses Webdriver to automate browser actions.
 
@@ -48,7 +48,7 @@ Cypress, because it meets all the requirements.
 
 * Good, because it is lightweight, no additional driver or dependencies required.
 * Good, because it is maintained.
-* Good, because it is easy to setup and use. 
+* Good, because it is easy to setup and use.
 * There are some drawbacks of Cypress but they're irrelevant to our use case:
     * No support for multiple browser tabs.
     * Doesn't support iFrame testing.
@@ -56,11 +56,11 @@ Cypress, because it meets all the requirements.
 
 ### TestCafe
 
-[TestCafe](https://testcafe.io/) is also a Javascript-based UI framework, and also operates straight in the browser. The only slight disadvantage it has as compared to Cypress is that it has a smaller community supporting/maintaining it. 
+[TestCafe](https://testcafe.io/) is also a Javascript-based UI framework, and also operates straight in the browser. The only slight disadvantage it has as compared to Cypress is that it has a smaller community supporting/maintaining it.
 
 * Good, because it is lightweight, no additional driver or dependencies required.
 * Good, because it is easy to setup and use.
-* Good, because it supports multiple tabs and iFrames. 
+* Good, because it supports multiple tabs and iFrames.
 * Bad, because it has a smaller community of maintainers as compared to Cypress.
 * Bad, because browsers are not aware that they are running in test mode. So, in some edge cases, automation control can be disrupted.
 
--- a/docs/design/adrs/2024-04-23-implement-prompflow-orchestration.md
+++ b/docs/design/adrs/2024-04-23-implement-prompflow-orchestration.md
@@ -1,6 +1,6 @@
 # Implement a PromptFlow Orchestration as an alternative for the chat backend
 
-* **Status:** proposed
+* **Status:** approved
 * **Proposer:** @superhindupur
 * **Date:** 2024-04-23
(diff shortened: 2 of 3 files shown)


docs/design/adrs/2024-03-18-use-cypres-as-a-framework-for-e2e-testing.md
```diff
@@ -1,17 +1,17 @@
 # Use Cypress as a framework for E2E testing
 
-* **Status:** proposed
+* **Status:** approved
 * **Proposer:** @superhindupur
```
superhindupur, line 4: Sorry for the many changes in this file, they're all whitespace changes apart from marking the ADR as approved

docs/design/adrs/2024-04-23-implement-prompflow-orchestration.md
```diff
@@ -0,0 +1,42 @@
+# Implement a PromptFlow Orchestration as an alternative for the chat backend
```
adamdougal, line 1: The title of the document suggests you think we **do** want to implement this. 
```suggestion
# Do not implement a PromptFlow Orchestration as an alternative for the chat backend
```
Would need ot rename the file as well

docs/design/adrs/2024-04-23-implement-prompflow-orchestration.md
```diff
@@ -0,0 +1,42 @@
...
+
+
+## Pros and Cons of the Options
+
+### Adding the PromptFlow orchestration
+* Good, because customers wanting to deploy the chat backend as a PromptFlow endpoint will be able to do so.
+* Good, because customers will have a low-code mechanism to experiment with the orchestration of the chat backend.
+* Bad, because building a PromptFlow orchestration that mimics the current chat backend requires significant effort - and it is unclear how much customer demand we have for it.
+* Bad, because the PromptFlow configuration can get out of date with the other orchestrators that are in code. This will cause confusion to customers and will lead them to open issues against the repo.
+* Bad, because outside of end-to-end tests that test an actual deployment, there is no way to test the PromptFlow configuration (i.e., no way to unit-test it).
```
adamdougal, line 36: I'd argue this is a benefit. Because it's just configuration, we won't need to test it. Any code based elements can still be unit tested.

docs/design/adrs/2024-04-23-implement-prompflow-orchestration.md
```diff
@@ -0,0 +1,42 @@
...
+
+
+### Not adding the PromptFlow Orchestration
+* Bad, because customers wanting to deploy the chat backend as a PromptFlow endpoint will NOT be able to do so.
+* Bad, because in the absence of a low-code PromptFlow alternative, customers will need coding knowledge to experiment with the chat backend orchestration.
+* Good, because there's one less thing to maintain / explain / test in the repo. :)
```
adamdougal, line 42: Could we put the good at the top of the list? My OCD is triggering a bit here 🙈

docs/design/adrs/2024-04-23-implement-prompflow-orchestration.md
```diff
@@ -0,0 +1,42 @@
...
+In the absence of concrete business / customer requirements, the outcome is to NOT add a PromptFlow orchestrator alternative for the chat backend.
+
+
+## Pros and Cons of the Options
+
+### Adding the PromptFlow orchestration
+* Good, because customers wanting to deploy the chat backend as a PromptFlow endpoint will be able to do so.
+* Good, because customers will have a low-code mechanism to experiment with the orchestration of the chat backend.
+* Bad, because building a PromptFlow orchestration that mimics the current chat backend requires significant effort - and it is unclear how much customer demand we have for it.
+* Bad, because the PromptFlow configuration can get out of date with the other orchestrators that are in code. This will cause confusion to customers and will lead them to open issues against the repo.
```
adamdougal, line 35: This is the same for langchain vs openai functions vs semantic kernel. Do you think we should align on just one orchestrator as a favourite?

docs/design/adrs/2024-04-23-implement-prompflow-orchestration.md
```diff
@@ -0,0 +1,42 @@
...
+## Decision Outcome
+In the absence of concrete business / customer requirements, the outcome is to NOT add a PromptFlow orchestrator alternative for the chat backend.
+
+
+## Pros and Cons of the Options
+
+### Adding the PromptFlow orchestration
+* Good, because customers wanting to deploy the chat backend as a PromptFlow endpoint will be able to do so.
+* Good, because customers will have a low-code mechanism to experiment with the orchestration of the chat backend.
+* Bad, because building a PromptFlow orchestration that mimics the current chat backend requires significant effort - and it is unclear how much customer demand we have for it.
```
adamdougal, line 34: I might be being optimistic, but I don't think it will be significant effort. Is it work breaking it down and doing rough estimates?

docs/design/adrs/2024-04-23-implement-prompflow-orchestration.md
superhindupur: Isn't semantic kernel a code-based orchestrator? If that's the case wouldn't the interfaces in code ensure that the functionality will remain common across it and other code-based orchestrators? If we don't have interfaces already we at least have the possibility of creating them and enforcing uniformity if they're in code.

With PF, that's not an option because the orchestration is in a dag file.

docs/design/adrs/2024-04-23-implement-prompflow-orchestration.md
superhindupur: > Because it's just configuration, we won't need to test it

e2e behavior will need to be tested for correctness, ordering of actions etc.

docs/design/adrs/2024-04-23-implement-prompflow-orchestration.md
superhindupur: Fair enough, the "significant effort" is subjective. My suggestion is - let's hold off on this until we're able to validate customer requirements with Randy tomorrow and maybe we will have a strong argument in favor of adding promptflow.

docs/design/adrs/2024-04-23-implement-prompflow-orchestration.md
adamdougal: Yea, but the negative of there being no way to test it, isn't a negative if you don't need to test it.

docs/design/adrs/2024-04-23-implement-prompflow-orchestration.md
adamdougal: > Isn't semantic kernel a code-based orchestrator?

Yea correct. There's only one common method which is:

```python
    @abstractmethod
    def orchestrate(
        self, user_message: str, chat_history: List[dict], **kwargs: dict
    ) -> dict:
        pass
```
https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/3e71fe10c50a129a20c314ba015065acdc6d8379/code/backend/batch/utilities/orchestrator/OrchestratorBase.py#L27-L31

Each orchestrator is then free to do what it likes within that. 

Obviously it's extra effort, but we should ensure as reviewers all changes made and features added are applied to all orchestration types.

docs/design/adrs/2024-04-23-implement-prompflow-orchestration.md
ross-p-smith: How about `2024-04-23-prompflow-orchestration.md`

superhindupur: Not applicable anymore since the decision matches with the file name.

docs/design/adrs/2024-04-23-implement-prompflow-orchestration.md
superhindupur: Fair enough, took this into consideration while making the final decision.

docs/design/adrs/2024-04-23-implement-prompflow-orchestration.md
superhindupur: I disagree with the point that we don't need to test it.

BUT the lack of e2e testing applies to all orchestrators at the moment. And when e2e tests are added, they can quite easily cover the promptflow backend too. Incorporated this into the final decision.

docs/design/adrs/2024-04-23-implement-prompflow-orchestration.md
```diff
@@ -0,0 +1,44 @@
+# Implement a PromptFlow Orchestration as an alternative for the chat backend
+
+* **Status:** proposed
```
adamdougal, line 3:
```suggestion
* **Status:** approved
```

adamdougal (approved):

adamdougal (approved):

PR #741, opened 2024-04-23
IV changes for Explore Data & Delete Data page

komalg1:
## Purpose

* Required by #321 
* This PR updates Explore & Delete pages on the Admin website corresponding to the Integrated Vectorization changes.
* Decouples the Explore & Delete page from the Search functionality.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
If 'AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION' is True, the Explore & Delete page should show the corresponding Ingested documents. 
No changes in the UI.
```


superhindupur:
The search client logic extraction is looking great - one thing I missed when we discussed this earlier, sorry - is it possible to create the clients from a central place so that the Explore and Delete pages don't have to worry about `if env_helper.AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION`? But not a big deal if this can't be done, just a thought.

Second - is it possible to add unit tests for the extracted search methods in the two clients, now that they're not connected with StreamLit directly anymore?

code/backend/pages/02_Explore_Data.py
```diff
@@ -1,13 +1,16 @@
...
+from batch.utilities.helpers.EnvHelper import EnvHelper
+from backend.pages.utilities.IntegratedVectorizationSearchHandler import (
+    IntegratedVectorizationSearchHandler,
+)
+from backend.pages.utilities.AzureSearchHandler import AzureSearchHandler
 
 sys.path.append(os.path.join(os.path.dirname(__file__), ".."))
+env_helper: EnvHelper = EnvHelper()
 
 load_dotenv()
```
ross-p-smith, line 15: I know not your actual change - but I wonder if this `load-dotenv()` is even used anymore? I would have thought it should have been before we load the envhelper

code/backend/pages/03_Delete_Data.py
```diff
@@ -2,10 +2,15 @@
...
+from batch.utilities.helpers.EnvHelper import EnvHelper
+from backend.pages.utilities.IntegratedVectorizationSearchHandler import (
+    IntegratedVectorizationSearchHandler,
+)
+from backend.pages.utilities.AzureSearchHandler import AzureSearchHandler
 
 sys.path.append(os.path.join(os.path.dirname(__file__), ".."))
+env_helper: EnvHelper = EnvHelper()
 
 load_dotenv()
```
ross-p-smith, line 15: Same with this one

komalg1:
> The search client logic extraction is looking great - one thing I missed when we discussed this earlier, sorry - is it possible to create the clients from a central place so that the Explore and Delete pages don't have to worry about `if env_helper.AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION`? But not a big deal if this can't be done, just a thought.
> 
> Second - is it possible to add unit tests for the extracted search methods in the two clients, now that they're not connected with StreamLit directly anymore?

Thanks for the review. Will take it up as a refactoring task & see if it is possible to have clients at a central place.
Added unit tests

code/backend/pages/02_Explore_Data.py
adamdougal: `load_dotenv()` actually happens within the `EnvHelper` too. So probably not required here at all.

code/backend/pages/utilities/AzureSearchHandler.py
```diff
@@ -0,0 +1,66 @@
+from backend.pages.utilities.SearchHandlerBase import SearchHandlerBase
+from backend.batch.utilities.helpers.AzureSearchHelper import AzureSearchHelper
+import json
+import pandas as pd
+import streamlit as st
```
adamdougal, line 5: I think _ideally_ we should try and avoild streamlit (and pandas?) from leaking outside of the pages. It makes this harder to reuse outside of the admin app.

Could we instead do something like create a `SearchResults` object that `perform_search` returns? It could then map what the search client returns to generic fields.

WDYT?

code/tests/pages_utilities/test_AzureSearchHandler.py
```diff
@@ -0,0 +1,70 @@
...
+        "backend.pages.utilities.AzureSearchHandler.st.stop"
+    ) as mock_stop:
+
+        # when
+        handler.delete_files(files)
+
+        # then
+        mock_success.assert_called_once_with("Deleted files: ['file1']")
+        assert not mock_info.called
+        assert not mock_stop.called
```
adamdougal, line 70: Is it worth verifying that we called `search_client.delete_documents` too?

code/backend/pages/utilities/AzureSearchHandler.py
```diff
@@ -0,0 +1,66 @@
...
+                files_to_delete.append(filename)
+                ids_to_delete += [{"id": id} for id in ids]
+
+        if len(ids_to_delete) == 0:
+            st.info("No files selected")
+            st.stop()
+
+        self.search_client.delete_documents(ids_to_delete)
+
+        st.success("Deleted files: " + str(files_to_delete))
```
adamdougal, line 66: Another reason why it's not ideal that we're invoking streamlit here, is that we need to add deleting from blob storage to this too in #448. Now that all the UI updating is in the `AzureSearchHandler` it's going to be awkward to get in.

code/tests/pages_utilities/test_AzureSearchHandler.py
adamdougal: Same for Integrated Vectorization!

code/backend/pages/utilities/SearchHandlerBase.py
```diff
@@ -0,0 +1,31 @@
+from abc import ABC, abstractmethod
+
+
+class SearchHandlerBase(ABC):
```
adamdougal, line 4: Love this abstraction!

adamdougal (changes requested):

code/tests/pages_utilities/test_AzureSearchHandler.py
komalg1: Updated

code/backend/pages/02_Explore_Data.py
komalg1: Removed load_dotenv

code/backend/pages/03_Delete_Data.py
```diff
@@ -35,59 +39,44 @@
...
-    search_client = vector_store_helper.get_vector_store().client
-
-    results = get_files()
-    files = output_results(results)
+    files = search_handler.output_results(results)
+    selections = {
+        filename: st.checkbox(filename, False, key=filename)
+        for filename in files.keys()
+    }
+    selected_files = {
```
adamdougal, line 64: Does this get updated as the user selects each item? Asking as it looks like this code is only invoked on loading, but it might be some streamlit magic making it work! 😅

code/backend/pages/01_Ingest_Data.py
```diff
@@ -145,7 +145,7 @@ def upload_file(bytes_data: bytes, file_name: str, content_type: Optional[str] =
             bytes_data,
             overwrite=True,
             content_settings=ContentSettings(content_type=content_type + charset),
-            metadata={"source_url": file_name},
+            metadata={"title": file_name},
```
ross-p-smith, line 148: I'm nervous that this is effectively a breaking change (unless I have misunderstood)

code/backend/pages/utilities/AzureSearchHandler.py
ross-p-smith, line 1: Love the fact that you have pulled these out of the streamlit app - would they be better being put in the `batch\utilities\search` folder rather than under `pages`?

code/backend/pages/01_Ingest_Data.py
komalg1: This should not be a breaking change. This was added in my last PR. Before that in push model this field is never used. I am updating it so that I am able to handle the '/' problem when pushing URL to blob storage. We can discuss more on this

code/backend/pages/03_Delete_Data.py
komalg1: Streamlit magic 🪄  It checks on every checkbox click

code/backend/pages/utilities/AzureSearchHandler.py
```diff
@@ -0,0 +1,49 @@
...
+        files = {}
+        for result in results:
+            id = result["id"]
+            filename = result["title"]
+            if filename in files:
+                files[filename].append(id)
+            else:
+                files[filename] = [id]
+
+        return files
```
superhindupur, line 38: Are we missing tests for output_results?

code/backend/pages/utilities/AzureSearchHandler.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def process_results(self, results):
+        data = [
+            [json.loads(result["metadata"])["chunk"], result["content"]]
+            for result in results
+        ]
+        return data
+
+    def get_files(self):
+        return self.search_client.search(
```
superhindupur, line 24: Missed test case for the get_files method here and in the IV search handler.

superhindupur:
Apart from some missed test cases, looks good to me. Once the tests are added I'm happy to approve. Thanks Komal!

superhindupur (approved):

cecheta (approved):

adamdougal (approved):
Great work!

PR #745, opened 2024-04-23
Enforced LF line endings

### LINKED ISSUE #742: scripts/package_frontend.sh causes build error because of Windows-style CRLF line endings (^M)
# Describe the bug
The script "package_frontend.sh" contains Windows-style CRLF line endings (^M), which causes issues when running the scripts in a UNIX/Linux environment like the Dev. Container

# Fix
I ran the following command on the package_frontend.sh: 
sed -i 's/\r$//' package_frontend.sh

# Debugging information
Run app in Dev Container

## Steps to reproduce
1. Build Dev. Container
2. Run azd up
3. Packaging service web failes

## Logs
```
vscode ➜ /workspaces/chat-with-your-data-solution-accelerator $ azd up

Packaging services (azd package)

  (✓) Done: Packaging service adminweb
  - Package Output: /tmp/chat-with-your-data-solution-accelerator-adminweb-azddeploy-1713871762.zip

  (✓) Done: Packaging service function
  - Package Output: /tmp/chat-with-your-data-solution-accelerator-function-azddeploy-1713871794.zip

  |       | Packaging service web/usr/bin/env: ‘bash\r’: No such file or directory
  (x) Failed: Packaging service web

ERROR: failed packaging service 'web': failed invoking event handlers for 'prepackage', 'prepackage' hook failed with exit code: '127', Path: '../scripts/package_frontend.sh'. : exit code: 127

ERROR: error executing step command 'package --all': failed packaging service 'web': failed invoking event handlers for 'prepackage', 'prepackage' hook failed with exit code: '127', Path: '../scripts/package_frontend.sh'. : exit code: 127
```
If applicable, add logs to help the engineer debug the problem.

---

# Tasks

- [ ] Task 1
- [ ] Task 2
- [ ] ...

ross-p-smith:
A user has reported that they are getting CRLF on a bash file. I do not see this on mine, but in order to fix #742 we can enforce line endings from git.

Whilst there I noticed that `make lint` was throwing errors so I have excluded the `.venv` directory as this is a generated folder. I have also removed the extra `build-frontend` that was being called in `ci`

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+* text=auto eol=lf
+*.{cmd,[cC][mM][dD]} text eol=crlf
+*.{bat,[bB][aA][tT]} text eol=crlf
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,4 @@
 [flake8]
 max-line-length = 88
 extend-ignore = E203, E501
+exclude = .venv
--- a/Makefile
+++ b/Makefile
@@ -20,11 +20,15 @@ include $(AZURE_ENV_FILE)
 help: ## 💬 This help message :)
 	@grep -E '[a-zA-Z_-]+:.*?## .*$$' $(firstword $(MAKEFILE_LIST)) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-23s\033[0m %s\n", $$1, $$2}'
 
-ci: lint unittest functionaltest build-frontend unittest-frontend ## 🚀 Continuous Integration (called by Github Actions)
+ci: lint unittest functionaltest frontendtest ## 🚀 Continuous Integration (called by Github Actions)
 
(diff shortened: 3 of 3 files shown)


.gitattributes
```diff
@@ -0,0 +1,3 @@
+* text=auto eol=lf
+*.{cmd,[cC][mM][dD]} text eol=crlf
+*.{bat,[bB][aA][tT]} text eol=crlf
```
cecheta, line 3: Powershell as well?

ross-p-smith: everything is `lf` - I believe that the only thing that needs `cr` is windows batch files

Based on https://code.visualstudio.com/docs/devcontainers/tips-and-tricks#_resolving-git-line-ending-issues-in-containers-resulting-in-many-modified-files

cecheta (approved):

Issue #761, opened 2024-04-26
Update IV app settings in Admin, Web & function


komalg1:
Required by #321 
# Description

App settings for Admin, Web & Backend function need to be updated to be able to run IV with feature flag enabled.

The settings that need to be added are:
AZURE_SEARCH_DATASOURCE_NAME
AZURE_SEARCH_INDEXER_NAME 
AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION - add this to the main.bicepparam file as well

Update LOCAL_DEPLOYMENT.md with these params

Update README.md with general info about IV

PR #767, opened 2024-04-29
Question Answer Tool changes

komalg1:
## Purpose

* Related to #321 
* This PR updates QuestionAnswerTool.py to work with Integrated Vectorization. It adds a new field 'chunk_id' to the lang chain SourceDocument.py class to be able to map it in UI.
* Added corresponding unit tests

```
[x] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
The Web chat application should be able to return results along with citations for the ingested documents.
```


code/backend/batch/utilities/helpers/EnvHelper.py
```diff
@@ -79,10 +79,10 @@ def __load_config(self, **kwargs) -> None:
             "AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION", "False"
         )
         self.AZURE_SEARCH_IV_MAX_PAGE_LENGTH = int(
-            os.getenv("AZURE_SEARCH_IV_MAX_PAGE_LENGTH", "2000")
+            os.getenv("AZURE_SEARCH_IV_MAX_PAGE_LENGTH", "800")
```
komalg1, line 82: Updated the default values because of token limit issue error

code/backend/batch/utilities/tools/QuestionAnswerTool.py
```diff
@@ -128,13 +139,43 @@ def generate_on_your_data_llm_chain(
             "chat_history": chat_history,
         }
 
+    def generate_source_documents(self, search_results):
```
komalg1, line 142: This method adds sas token placeholder for the documents in Storage so that the link from citation redirects to the correct blob storage

code/backend/batch/utilities/helpers/EnvHelper.py
```diff
@@ -79,10 +79,10 @@ def __load_config(self, **kwargs) -> None:
...
         )
         self.AZURE_SEARCH_IV_PAGE_OVERLAP_LENGTH = int(
-            os.getenv("AZURE_SEARCH_IV_PAGE_OVERLAP_LENGTH", "500")
```
ross-p-smith, line 85: Can we also change the .env.sample (even though I'm unsure its used!) and in the LOCAL_DEPLOYMENT.md I dont think these new values have been added.

code/backend/batch/utilities/tools/QuestionAnswerTool.py
ross-p-smith: Why not put that comment in as a remark for the `generate_source_documents` method?

code/backend/batch/utilities/helpers/EnvHelper.py
komalg1: I will update all the .md files, .env.sample in a separate PR. Will do it along with #761

code/backend/batch/utilities/parser/OutputParserTool.py
```diff
@@ -68,7 +68,11 @@ def parse(
                 {
                     "content": doc.get_markdown_url() + "\n\n\n" + doc.content,
                     "id": doc.id,
-                    "chunk_id": doc.chunk,
+                    "chunk_id": (
+                        re.findall(r"\d+", doc.chunk_id)[-1]
+                        if doc.chunk_id is not None
+                        else doc.chunk
+                    ),
```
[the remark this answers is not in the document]
komalg1: Added

code/backend/batch/utilities/tools/QuestionAnswerTool.py
```diff
@@ -128,13 +139,43 @@ def generate_on_your_data_llm_chain(
...
+        sources = []
+        for result in search_results:
+            original_source = result.get("source", "")
+
+            # Find all occurrences of 'http' or 'https'
+            matches = list(re.finditer(r"https?://", original_source))
+
+            if len(matches) > 1:
```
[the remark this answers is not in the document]
komalg1: Done

code/backend/batch/utilities/tools/QuestionAnswerTool.py
komalg1: Moved it into a separate method as per below comment from Adam

code/backend/batch/utilities/tools/QuestionAnswerTool.py
```diff
@@ -128,13 +139,40 @@ def generate_on_your_data_llm_chain(
...
+            }
+            sources.append(
+                Document(
+                    page_content=result["content"],
+                    metadata=metadata_dict,
+                )
+            )
+        return sources
+
+    def _find_all_http_or_https(self, original_source):
```
adamdougal, line 161: Sorry I think I confused this by suggesting to extract a small part and then all of it! This is doing more than just the http/https part so maybe should be called something like 
```suggestion
    def _extract_source_url(self, original_source):
```

code/backend/batch/utilities/tools/QuestionAnswerTool.py
```diff
@@ -29,9 +33,16 @@
 class QuestionAnswerTool(AnsweringToolBase):
     def __init__(self) -> None:
         self.name = "QuestionAnswer"
-        self.vector_store = AzureSearchHelper().get_vector_store()
-        self.verbose = True
         self.env_helper = EnvHelper()
+        self.llm_helper = LLMHelper()
+        if self.env_helper.AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION:
```
ross-p-smith, line 38: We have this code block a few times in the codebase, wondering whether a factory method would be better would also help testing

code/backend/batch/utilities/parser/OutputParserTool.py
```diff
@@ -68,7 +68,11 @@ def parse(
                 {
                     "content": doc.get_markdown_url() + "\n\n\n" + doc.content,
                     "id": doc.id,
-                    "chunk_id": doc.chunk,
+                    "chunk_id": (
+                        re.findall(r"\d+", doc.chunk_id)[-1]
```
cecheta, line 72: Is `re.findall(r"\d+", doc.chunk_id)` guaranteed to have at least one item?

komalg1: So with IV, chunk_id will be in the form of "guid_page_1". 1 here is the chunk, so `re.findall(r"\d+", doc.chunk_id)` should return the last digit

code/backend/batch/utilities/tools/QuestionAnswerTool.py
```diff
@@ -161,15 +191,24 @@ def answer_question(self, question: str, chat_history: list[dict], **kwargs: dic
...
-            source_document = SourceDocument(
-                id=source.metadata["id"],
-                content=source.page_content,
-                title=source.metadata["title"],
-                source=source.metadata["source"],
-                chunk=source.metadata["chunk"],
-                offset=source.metadata["offset"],
-                page_number=source.metadata["page_number"],
-            )
+            if self.env_helper.AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION:
```
ross-p-smith, line 194: I feel like this could be a static method on SourceDocument. e.g. `source_document = SourceDocument.Create(source=source, use_integrated_vectorisation=self.env_helper.AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION)` or something similar. Interested in others views?

code/backend/batch/utilities/tools/QuestionAnswerTool.py
```diff
@@ -128,13 +131,40 @@ def generate_on_your_data_llm_chain(
...
+        return source_url
+
     def answer_question(self, question: str, chat_history: list[dict], **kwargs: dict):
-        # Retrieve documents as sources
-        sources = self.vector_store.similarity_search(
-            query=question,
-            k=self.env_helper.AZURE_SEARCH_TOP_K,
-            filters=self.env_helper.AZURE_SEARCH_FILTER,
-        )
+        if self.env_helper.AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION:
```
ross-p-smith, line 163: Now you have the code/backend/batch/utilities/search/Search.py file, I wonder if a new method `retrieve_documents_as_sources` could live there and have this block of code too

code/backend/batch/utilities/tools/QuestionAnswerTool.py
superhindupur: Yes, if we have two flavors of SourceDocument, like we do with the AzureSearchHandler - one for IV and one for push - we can keep all the IV/non-IV handling of the source documents there. I would even move the generate_source_documents method there or under the search handler. 

Basically, QuestionAnswerTool should have as less details of what to do if AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION is true as possible.

code/backend/batch/utilities/search/Search.py
```diff
@@ -0,0 +1,13 @@
+from ..search.AzureSearchHandler import AzureSearchHandler
+from ..search.IntegratedVectorizationSearchHandler import (
+    IntegratedVectorizationSearchHandler,
+)
+
+
+class Search:
```
superhindupur, line 7: I really like this. I think in general we should hide all the IV/non-IV if-else logic behind factory classes and handlers like you did here.

code/backend/batch/utilities/tools/QuestionAnswerTool.py
komalg1: If we have 2 SourceDocuments, tomorrow if we add new fields, do we create another type of SourceDocument?
If we use something like `SourceDocument.Create` , won't there be a lot of code duplication because all the methods of SourceDocument will be updated?

superhindupur:
LGTM too. Will wait for you to fix the unit tests, and I can approve. Thanks Komal!

code/backend/batch/utilities/search/Search.py
```diff
@@ -2,6 +2,8 @@
 from ..search.IntegratedVectorizationSearchHandler import (
     IntegratedVectorizationSearchHandler,
 )
+from langchain_core.documents import Document
+import re
```
superhindupur, line 6: Is this an unused import?

superhindupur (approved):

Issue #736, opened 2024-04-23
Add testing for the admin app pages.


komalg1:
- Setup testing framework for Admin website which uses Streamlit module.
- Test the business logic in the admin app.

PR #782, opened 2024-04-29
Remove duplicate upload_file logic from Admin Ingest page.

superhindupur:
## Purpose
While looking at setting up tests for the admin app pages under #736, I realized that we can refactor the pages to have a bare minimum of business logic in them, with the logic coming from the backend utilities package.

In this PR, I've removed the dupliocate upload_file to blob storage functionality and used the functionality from the AzureBlobStorageHelper.py file instead. I've also added a missing unit test to cover this functionality, achieving the overall goal of testing the admin app pages better.

```
[ ] Yes
[x ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x ] Refactoring (no functional changes, no api changes)
[ ] Other... Please describe:
```

## How to Test
Deploy the admin app and test whether the following are working:
* Ingestion of files
* Ingestion of URLs 
* I've tested the above with and without integrated vectorization, and for txt and pdf files.


komalg1:
Was this tested with Integrated Vectorization enabled for URLs?

code/backend/batch/utilities/helpers/AzureBlobStorageHelper.py
```diff
@@ -101,16 +103,26 @@ def file_exists(self, file_name):
...
         return blob_client.exists()
 
-    def upload_file(self, bytes_data, file_name, content_type="application/pdf"):
+    def upload_file(self, bytes_data, file_name, content_type: Optional[str] = None):
         # Create a blob client using the local file name as the name for the blob
         blob_client = self.blob_service_client.get_blob_client(
             container=self.container_name, blob=file_name
         )
+
+        if content_type is None:
```
adamdougal, line 112: Can you add a test where this is **not** none. I think it would currently leave the `charset` as `None` and I'm not sure what python does when you try to add `None` to a string.

ross-p-smith (commented):
Thank you for doing this, that's a nice chunk of code to delete. Is it to much to align the name of the file/class name in this change (AzureBlobStorageHelper(filename)/AzureBlobStorageClient(classname))

code/backend/batch/utilities/helpers/AzureBlobStorageHelper.py
superhindupur: Good catch! Fixed and added a test

code/backend/batch/utilities/helpers/AzureBlobStorageHelper.py
```diff
@@ -101,16 +103,26 @@ def file_exists(self, file_name):
...
+                else ""
+            )
+            content_type = content_type if content_type is not None else "text/plain"
+
         # Upload the created file
         blob_client.upload_blob(
             bytes_data,
             overwrite=True,
-            content_settings=ContentSettings(content_type=content_type),
+            content_settings=ContentSettings(content_type=content_type + charset),
```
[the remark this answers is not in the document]
superhindupur: Thanks for pointing this out, added, and covered with a test

code/tests/utilities/test_AzureBlobStorageClient.py
```diff
@@ -46,6 +46,51 @@ def test_file_exists(BlobServiceClientMock: MagicMock, exists: bool, expected: b
...
+    blob_client_mock.upload_blob.assert_called_once_with(
+        str.encode("mock-data"),
+        overwrite=True,
+        content_settings=ANY,
+        metadata={"title": "mock-file"},
+    )
+    _, kwargs = blob_client_mock.upload_blob.call_args
+    assert (
+        kwargs["content_settings"]["content_type"] == content_type
+        if content_type is not None
```
adamdougal, line 77: I try to avoid if conditions in tests where possible and keep any logic to a minimum. It tends to make them harder to follow and maintain.

Could you use a tuple on line 51 instead to also pass in the expectation? You can see an example here https://docs.pytest.org/en/7.1.x/how-to/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions

code/backend/batch/utilities/helpers/AzureBlobStorageClient.py
```diff
@@ -101,16 +103,30 @@ def file_exists(self, file_name):
...
+            content_type = content_type if content_type is not None else "text/plain"
+            content_settings = ContentSettings(content_type=content_type + charset)
+
         # Upload the created file
         blob_client.upload_blob(
             bytes_data,
             overwrite=True,
-            content_settings=ContentSettings(content_type=content_type),
+            content_settings=content_settings,
+            metadata={"title": file_name},
```
cecheta, line 129: I would think it would be better to make `metadata` an optional parameter and pass it in from where this function is being called in the Admin app, because this function is used in multiple places and we might not always want metadata

code/backend/pages/01_Ingest_Data.py
```diff
@@ -78,21 +53,24 @@ def remote_convert_files_and_add_embeddings(process_all=False):
...
+    blob_client: AzureBlobStorageClient, urls: list[str]
+):
     for url in urls:
         try:
             response = requests.get(url)
             parsed_data = BeautifulSoup(response.content, "html.parser")
             with io.BytesIO(parsed_data.get_text().encode("utf-8")) as stream:
-                upload_file(stream, url)
+                st.session_state["filename"] = url
+                st.session_state["file_url"] = blob_client.upload_file(stream, url)
```
cecheta, line 73: Need `metadata={"title": file_name}` here

code/backend/pages/01_Ingest_Data.py
```diff
@@ -207,13 +105,17 @@ def upload_file(bytes_data: bytes, file_name: str, content_type: Optional[str] =
...
         if uploaded_files is not None:
             for up in uploaded_files:
                 # To read file as bytes:
                 bytes_data = up.getvalue()
                 if st.session_state.get("filename", "") != up.name:
                     # Upload a new file
-                    upload_file(bytes_data, up.name)
+                    st.session_state["filename"] = up.name
+                    st.session_state["file_url"] = blob_client.upload_file(
+                        bytes_data, up.name
```
cecheta, line 117: and here

adamdougal (dismissed):
LGTM!

cecheta (dismissed):

ross-p-smith (approved):
LGTM

PR #783, opened 2024-04-29
Removing transformers as not used

ross-p-smith:
## Purpose
Relates to #417 
Remove 2 packages that we are not using
Also upgraded a few packages

## How to Test
- Was able to run all tests
- Ingested a  document
- Asked a question on the document


adamdougal (approved):
lgtm

Issue #748, opened 2024-04-24
Generate and store image embeddings in search index


adamdougal:
Required by https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/issues/323

# Description

Generate image embeddings using computer vision and store them in a modified search index.

# Tasks

- [x] Investigate what needs to change with the search index and how to encorporate that into CWYDSA
  - [x] POC: https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/786
  - [x] ADR: https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/788
- [x] Add functional test for current index creation - https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/802
- [x] Stop using Langchain for creating, populating and searching index - https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/827
- [x] Generate image embeddings using computer vision - https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/892
- [x] Store image embeddings in search index - https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/921

adamdougal:
**Update 26th April:**

- I'm investigating what changes are going to be required to the app code and index to be able to store and query an additional image embedding
- I've hit a problem where it does not appear that LangChain has support for Azure AI Computer Vision OR has the ability to query AzureSearch for more than one vector embedding.
- One option that _might_ work is to have two separate indexes and combine the search results in the application code. I'm not 100% sure if this is viable as yet
- Another option is to remove the use of LangChain in the [QuestionAnswerTool ](https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/main/code/backend/batch/utilities/tools/QuestionAnswerTool.py)and [AzureSearchHelper](https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/main/code/backend/batch/utilities/helpers/AzureSearchHelper.py)
  - We would still keep the LangChain orchestration option, but it would only be used in the [LangChainAgent](https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/main/code/backend/batch/utilities/orchestrator/LangChainAgent.py)
  - This has also been done (for now) in the [azure-sample-openai-demo](https://github.com/Azure-Samples/azure-search-openai-demo/pull/738) repo
  - **Benefits:**
[...]

PR #786, opened 2024-04-30
POC: Stop using langchain for Azure Search and OpenAI calls

adamdougal:
- This is POC code, it's not unit tested and if we proceed it may be structured differently
- Retrains langchain as an overall orchestrator
- Motivation for doing this is to give us full control over the search index and how we query it, allowing us to add an image vector field

### DIFF
--- a/code/backend/batch/utilities/helpers/AzureSearchHelper.py
+++ b/code/backend/batch/utilities/helpers/AzureSearchHelper.py
@@ -1,3 +1,4 @@
+import logging
 from langchain.vectorstores.azuresearch import AzureSearch
 from azure.search.documents.indexes.models import (
     SearchableField,
@@ -7,6 +8,21 @@ from azure.search.documents.indexes.models import (
 )
 from .LLMHelper import LLMHelper
 from .EnvHelper import EnvHelper
+from azure.search.documents.indexes import SearchIndexClient
+from azure.search.documents import SearchClient
+from azure.core.credentials import AzureKeyCredential
+from azure.identity import DefaultAzureCredential
+from azure.search.documents.indexes.models import (
+    HnswAlgorithmConfiguration,
+    HnswParameters,
+    SearchIndex,
+    SemanticConfiguration,
+    SemanticField,
+    SemanticPrioritizedFields,
+    SemanticSearch,
+    VectorSearch,
+    VectorSearchProfile,
+)
 
 
 class AzureSearchHelper:
@@ -16,6 +32,24 @@ class AzureSearchHelper:
         self.llm_helper = LLMHelper()
         self.env_helper = EnvHelper()
 
+        if self.env_helper.is_auth_type_keys:
+            search_credential = AzureKeyCredential(self.env_helper.AZURE_SEARCH_KEY)
+        else:
+            search_credential = DefaultAzureCredential()
+
+        logging.info(
+            f"Creating search client for {self.env_helper.AZURE_SEARCH_SERVICE} and index {self.env_helper.AZURE_SEARCH_INDEX}"
+        )
+
+        self.search_client = SearchClient(
+            endpoint=self.env_helper.AZURE_SEARCH_SERVICE,
+            index_name=self.env_helper.AZURE_SEARCH_INDEX,
+            credential=search_credential,
+        )
+        self.search_index_client = SearchIndexClient(
+            endpoint=self.env_helper.AZURE_SEARCH_SERVICE, credential=search_credential
+        )
+
     @property
     def search_dimensions(self) -> int:
         if AzureSearchHelper._search_dimension is None:
@@ -24,7 +58,10 @@ class AzureSearchHelper:
             )
         return AzureSearchHelper._search_dimension
 
-    def get_vector_store(self):
+    def get_client(self):
+        return self.search_client
+
+    def create_or_update_index(self):
         fields = [
             SimpleField(
                 name="id",
@@ -70,24 +107,51 @@ class AzureSearchHelper:
             ),
         ]
 
-        return AzureSearch(
-            azure_search_endpoint=self.env_helper.AZURE_SEARCH_SERVICE,
-            azure_search_key=(
-                self.env_helper.AZURE_SEARCH_KEY
-                if self.env_helper.AZURE_AUTH_TYPE == "keys"
+        index = SearchIndex(
+            name=self.env_helper.AZURE_SEARCH_INDEX,
+            fields=fields,
+            semantic_search=(
+                SemanticSearch(
+                    configurations=[
+                        SemanticConfiguration(
+                            name=self.env_helper.AZURE_SEARCH_SEMANTIC_SEARCH_CONFIG,
+                            prioritized_fields=SemanticPrioritizedFields(
+                                title_field=None,
+                                content_fields=[SemanticField(field_name="content")],
+                            ),
+                        )
+                    ]
+                )
+                if self.env_helper.AZURE_SEARCH_USE_SEMANTIC_SEARCH
                 else None
             ),
-            index_name=self.env_helper.AZURE_SEARCH_INDEX,
-            embedding_function=self.llm_helper.get_embedding_model().embed_query,
-            fields=fields,
-            search_type=(
-                "semantic_hybrid"
-                if self.env_helper.AZURE_SEARCH_USE_SEMANTIC_SEARCH
-                else "hybrid"
+            vector_search=VectorSearch(
+                algorithms=[
+                    HnswAlgorithmConfiguration(
+                        name="hnsw_config",
+                        parameters=HnswParameters(metric="cosine"),
+                    )
+                ],
+                profiles=[
+                    VectorSearchProfile(
+                        name="myHnswProfile",
+                        algorithm_configuration_name="hnsw_config",
+                    ),
+                ],
             ),
-            semantic_configuration_name=self.env_helper.AZURE_SEARCH_SEMANTIC_SEARCH_CONFIG,
-            user_agent="langchain chatwithyourdata-sa",
         )
+
+        if self.env_helper.AZURE_SEARCH_INDEX not in [
+            name for name in self.search_index_client.list_index_names()
+        ]:
+            try:
+                logging.info(f"Creating index {self.env_helper.AZURE_SEARCH_INDEX}")
+                self.search_index_client.create_index(index)
+            except Exception as e:
+                logging.exception("Error Creating index")
+                raise e
+        else:
+            logging.info(f"Index {self.env_helper.AZURE_SEARCH_INDEX} already exists")
 
     def get_conversation_logger(self):
         fields = [
--- a/code/backend/batch/utilities/helpers/DocumentProcessorHelper.py
+++ b/code/backend/batch/utilities/helpers/DocumentProcessorHelper.py
@@ -42,11 +42,13 @@ class Processor(ChunkingSettings, LoadingSettings):
 
 class DocumentProcessor:
     def __init__(self):
-        pass
+        self.azure_search_helper = AzureSearchHelper()
+        self.search_client = self.azure_search_helper.search_client
+        self.llm_helper = LLMHelper()
 
     def process(self, source_url: str, processors: List[Processor]):
-        vector_store_helper = AzureSearchHelper()
-        vector_store = vector_store_helper.get_vector_store()
+        logger.info(f"Processing {source_url} with processors {processors}")
+        self.azure_search_helper.create_or_update_index()
         for processor in processors:
             if not processor.use_advanced_image_processing:
                 try:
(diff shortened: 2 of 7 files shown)


code/backend/batch/utilities/helpers/AzureSearchHelper.py
```diff
@@ -70,25 +107,52 @@ def get_vector_store(self):
...
-            azure_search_key=(
-                self.env_helper.AZURE_SEARCH_KEY
-                if self.env_helper.AZURE_AUTH_TYPE == "keys"
-                else None
-            ),
-            index_name=self.env_helper.AZURE_SEARCH_INDEX,
-            embedding_function=self.llm_helper.get_embedding_model().embed_query,
+        index = SearchIndex(
+            name=self.env_helper.AZURE_SEARCH_INDEX,
             fields=fields,
```
adamdougal, line 112: Note, the fields are unchanged. A new field will be added for image embeddings if enabled

code/backend/batch/utilities/helpers/DocumentProcessorHelper.py
```diff
@@ -55,12 +57,21 @@ def process(self, source_url: str, processors: List[Processor]):
...
-                    return vector_store.add_documents(documents=documents, keys=keys)
+
+                    for document in documents:
+                        doc = {
+                            "id": document.id,
+                            "content": document.content,
+                            "content_vector": self.llm_helper.generate_embeddings(
+                                document.content
+                            ),
+                            # "metadata": document.metadata, this is just a full json escaped version of the entry
```
adamdougal, line 68: Not really sure why this is needed. I can add this back later though.

code/backend/batch/utilities/tools/QuestionAnswerTool.py
```diff
@@ -1,182 +1,113 @@
...
-            )
-            source_documents.append(source_document)
+        response = self.llm_helper.get_chat_completion(messages)
+        logger.info(response)
 
-        clean_answer = Answer(
+        return Answer(
             question=question,
-            answer=answer,
+            answer=response.choices[0].message.content,
```
adamdougal, line 109: Would need to check if `choices` has any content etc.

code/backend/batch/utilities/tools/QuestionAnswerTool.py
```diff
@@ -1,182 +1,113 @@
...
-                MessagesPlaceholder("chat_history"),
-                HumanMessagePromptTemplate.from_template(
-                    self.config.prompts.answering_user_prompt
-                ),
+        if self.config.prompts.use_on_your_data_format:
+            example_question = self.config.prompts.answering_user_prompt.replace(
+                "{sources}", self.config.example.documents
+            ).replace("{question}", self.config.example.user_question)
+            messages = [
+                # {"role": "system", "content": f"{self.config.prompts.answering_system_prompt} {self.env_helper.AZURE_OPENAI_SYSTEM_MESSAGE}"}, # TODO is this second system message needed?
```
adamdougal, line 78: I'm not really sure why we have this second system message

code/backend/batch/utilities/tools/QuestionAnswerTool.py
```diff
@@ -1,182 +1,113 @@
...
 
-        sources_text = "\n\n".join(
-            [f"[doc{i+1}]: {source.page_content}" for i, source in enumerate(sources)]
+        # Now that we are querying Azure Search using the official libraries, we can add and modify the query any way we need, such as adding a query for image embeddings
+        vector_query = VectorizedQuery(
+            vector=question_embeddings,
+            k_nearest_neighbors=self.env_helper.AZURE_SEARCH_TOP_K,
+            fields="content_vector",
+        )
+        search_results = self.azure_search_helper.search_client.search(
```
adamdougal, line 39: This is the main reason for this change. It gives us full control of the query to AI Search, allowing us to supply multiple `VectorizedQuery`'s, one fo the text content and if enabled, one for the image content. We can then pass any image results to gpt-4 vision in the messages below

code/backend/batch/utilities/helpers/AzureSearchHelper.py
```diff
@@ -24,7 +58,10 @@ def search_dimensions(self) -> int:
             )
         return AzureSearchHelper._search_dimension
 
-    def get_vector_store(self):
+    def get_client(self):
+        return self.search_client
+
+    def create_or_update_index(self):
```
cecheta, line 64: This is named Create or Update but I don't think it updates?

code/backend/batch/utilities/helpers/DocumentProcessorHelper.py
```diff
@@ -55,12 +57,21 @@ def process(self, source_url: str, processors: List[Processor]):
...
+                            "title": document.title,
+                            "source": document.source,
+                            "chunk": document.chunk,
+                            "offset": document.offset,
+                        }
+                        self.search_client.upload_documents([doc])
```
cecheta, line 74: Could you upload all of these at once?

code/backend/batch/utilities/helpers/AzureSearchHelper.py
```diff
@@ -7,6 +8,21 @@
 )
 from .LLMHelper import LLMHelper
 from .EnvHelper import EnvHelper
+from azure.search.documents.indexes import SearchIndexClient
+from azure.search.documents import SearchClient
+from azure.core.credentials import AzureKeyCredential
+from azure.identity import DefaultAzureCredential
+from azure.search.documents.indexes.models import (
```
ross-p-smith, line 15: nit: Haven't we got an import above that uses `azure.search.documents.indexes.models`

code/backend/batch/utilities/helpers/AzureSearchHelper.py
```diff
@@ -1,3 +1,4 @@
+import logging
 from langchain.vectorstores.azuresearch import AzureSearch
```
ross-p-smith, line 2: Do we still need this import?

code/backend/batch/utilities/tools/PostPromptTool.py
```diff
@@ -14,53 +11,34 @@ def validate_answer(self, answer: Answer) -> dict:
...
         else:
             return Answer(
                 question=answer.question,
-                answer=answer.answer,
-                source_documents=answer.source_documents,
-                prompt_tokens=cb.prompt_tokens,
-                completion_tokens=cb.completion_tokens,
+                answer=config.messages.post_answering_filter,
+                source_documents=[],
+                prompt_tokens=response.prompt_tokens,
```
cecheta, line 42: We have `response.usage` above but just `response` here

code/backend/batch/utilities/tools/QuestionAnswerTool.py
cecheta: It was from the On Your Data prompt, it should come after the examples

code/backend/batch/utilities/helpers/DocumentProcessorHelper.py
```diff
@@ -42,11 +42,13 @@ def __eq__(self, other):
 
 class DocumentProcessor:
```
ross-p-smith, line 43: Shall we rename this file whilst here and get rid of that horrible non descript word "helper"? :-D

code/backend/batch/utilities/tools/QuestionAnswerTool.py
```diff
@@ -1,182 +1,113 @@
...
-                    self.config.prompts.answering_user_prompt
-                ),
-                AIMessagePromptTemplate.from_template("{answer}"),
-            ]
-        )
-
-        few_shot_prompt = FewShotChatMessagePromptTemplate(
-            example_prompt=example_prompt,
-            examples=examples,
+        ret_docs_json = json.dumps(
```
cecheta, line 66: The JSON sources is only used with the new format, but a different sources format is used with the old format, which is why it was separate before

code/backend/batch/utilities/helpers/DocumentProcessorHelper.py
ross-p-smith: I'm happy with leaving a link to a new issue to investigate later if ok with you - does seem odd to have an escaped version to pass around.

code/backend/batch/utilities/tools/QuestionAnswerTool.py
```diff
@@ -1,182 +1,113 @@
...
-        # Retrieve documents as sources
-        sources = self.vector_store.similarity_search(
-            query=question,
-            k=self.env_helper.AZURE_SEARCH_TOP_K,
-            filters=self.env_helper.AZURE_SEARCH_FILTER,
-        )
+            for message in chat_history:
+                messages.append(
+                    {"role": message["role"], "content": message["content"]}
+                )  # TODO should this be in the else too?
```
cecheta, line 94: When adding the new prompt, I didn't want to make any changes to the old prompt. The old prompt wasn't using chat history at all

code/backend/batch/utilities/tools/QuestionAnswerTool.py
```diff
@@ -1,182 +1,113 @@
...
-            [
-                SystemMessage(content=self.config.prompts.answering_system_prompt),
-                few_shot_prompt,
-                SystemMessage(content=self.env_helper.AZURE_OPENAI_SYSTEM_MESSAGE),
-                MessagesPlaceholder("chat_history"),
-                HumanMessagePromptTemplate.from_template(
-                    self.config.prompts.answering_user_prompt
-                ),
+        if self.config.prompts.use_on_your_data_format:
+            example_question = self.config.prompts.answering_user_prompt.replace(
```
cecheta, line 74: Perhaps it's because it's a POC, but we should probably keep the logic about whether to use the example or not

code/backend/batch/utilities/tools/QuestionAnswerTool.py
```diff
@@ -1,182 +1,113 @@
...
+                {
+                    "role": "system",
+                    "content": self.config.prompts.answering_system_prompt,
+                },
+                {"role": "system", "name": "example_user", "content": example_question},
+                {
+                    "role": "system",
+                    "name": "example_assistant",
+                    "content": self.config.example.answer,
+                },
```
cecheta, line 88: I don't think we need `name` here, and also the roles should be `system`, then `user`, then `assistant`

cecheta (commented):
Looking good, this should help us going forwards

code/backend/batch/utilities/tools/QuestionAnswerTool.py
```diff
@@ -1,182 +1,113 @@
...
 
 
 class QuestionAnswerTool(AnsweringToolBase):
+
     def __init__(self) -> None:
         self.name = "QuestionAnswer"
-        self.vector_store = AzureSearchHelper().get_vector_store()
-        self.verbose = True
+        self.azure_search_helper = AzureSearchHelper()
+        self.llm_helper = LLMHelper()
```
ross-p-smith, line 24: Assuming that #767 is going to conflict here

code/backend/pages/02_Explore_Data.py
```diff
@@ -40,21 +39,18 @@
...
     results = search_client.search(
-        "*", select="title, content, metadata", filter=f"title eq '{filename}'"
+        "*", select="title, content, chunk", filter=f"title eq '{filename}'"
     )
 
-    data = [
-        [json.loads(result["metadata"])["chunk"], result["content"]]
-        for result in results
-    ]
+    data = [[result["chunk"], result["content"]] for result in results]
```
ross-p-smith, line 53: so that metadata was just encoded to we could load it here?!?!

ross-p-smith:
I'm very happy with the direction of this - it actually simplifies things and removes a *lot* of langchain! Which will ultimately help @cecheta !! :-D I think we need Komal's in first though.

code/backend/batch/utilities/helpers/AzureSearchHelper.py
adamdougal: I'll verify this and rename accordingly when/if we do this for real

code/backend/batch/utilities/helpers/DocumentProcessorHelper.py
adamdougal: Great point, I'll do this and when/if we do this for real.

code/backend/batch/utilities/tools/QuestionAnswerTool.py
adamdougal: Wow I didn't realise that!

code/backend/batch/utilities/tools/QuestionAnswerTool.py
adamdougal: Ah yea I'll re-add that when/if we do this for real.

code/backend/batch/utilities/tools/QuestionAnswerTool.py
adamdougal: Yes it is, I'll let that go in first before making changes for real.

code/backend/pages/02_Explore_Data.py
adamdougal: This is one of the usages, it's used in a couple of other places, not sure why it needs to. I _probably_ won't remove it when/if I do this for real to keep changes to a minimum, I was just being lazy in this POC avoiding json.dumping it.

PR #788, opened 2024-04-30
Raise ADR proposing to remove Langchain from tools

adamdougal:
- To enable storing and querying image embeddings


ross-p-smith (approved):
Clear, to the point - makes perfect sense!

cecheta (approved):

docs/design/adrs/2024-04-30-remove-langchain-from-tools.md
```diff
@@ -0,0 +1,87 @@
...
+* Keep Langchain but not if advanced image processing enabled
+
+## Decision Outcome
+
+Remove Langchain from tools. This gives us full control to deliver this feature. It will also allow us to implement
+cutting edge features in future without having to wait for Langchain to be updated.
+
+## Pros and Cons of the Options
+
+### Remove Langchain from tools
```
superhindupur, line 45: This section is missing details on what we would lose out on if we removed langchain from "tools".
It would be beneficial to understand why the original contributors chose to include langchain in non-orchestration code.
It would also make sense to add a brief summary of what these tools are doing - is it the search flow? How is langchain contributing to it currently?

adamdougal: Good point! I've added a note in the bad section for this.

superhindupur (approved):

ross-p-smith (approved):

PR #798, opened 2024-05-01
[Integrated Vectorization] Changes for Admin Configuration page

### LINKED ISSUE #762: Updating Admin Config page to allow configuring the page length & page overlap & hide the layout chunking strategy part
Required by #321 

# Description

The Admin configuration page needs to allow configuring the Chunk size & Overlap size when using Integrated Vectorization.
Layout chunking strategy section needs to be hidden as it is not needed when using IV.

# Tasks

_To be filled in by the engineer picking up the subtask

- [x] Give option to user to configure chunk size & overlap size
- [x] Hide layout strategy section in Config page when AZURE_USE_INTEGRATED_VECTORIZATION is true

komalg1:
## Purpose

* Resolves #762 
* Related to #321
* This PR adds Max page length & overlap length config values that can be modified from Admin configuration page.
* Adds a factory method for Processing the ingested file.
* Added & updated unit tests

```
[x] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
If 'AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION' is true, user should be able to see two input values for Max page length & overlap length
```
image

## What to Check
* Any updates to the config values should be saved.


code/backend/batch/utilities/helpers/DocumentProcessorHelper.py
komalg1, line 1: Moved to different folder

code/backend/batch/utilities/helpers/Processor.py
komalg1, line 1: Moved the class out of DocumentProcessorHelper.py file

code/backend/batch/utilities/helpers/processors/DocumentProcessorHelper.py
```diff
@@ -0,0 +1,56 @@
...
+from .ProcessorBase import ProcessorBase
+from ..AzureSearchHelper import AzureSearchHelper
+from ..DocumentLoadingHelper import DocumentLoading
+from ..DocumentChunkingHelper import DocumentChunking
+from ...common.SourceDocument import SourceDocument
+
+logger = logging.getLogger(__name__)
+
+
+class DocumentProcessor(ProcessorBase):
```
ross-p-smith, line 18: Naming is hard, but I feel this should capture what it is "IndexProcessor" or "PushVectorizer" (I can't think of a good name!) - but DocumentProcessor is not it. Can we also remove the "Helper" from the filename.

Or Something like PushEmbedder?!?

code/backend/batch/utilities/helpers/processors/Process.py
```diff
@@ -0,0 +1,14 @@
+from ..AzureBlobStorageClient import AzureBlobStorageClient
+from .DocumentProcessorHelper import DocumentProcessor
+from .IntegratedVectorizationProcessorHelper import (
+    IntegratedVectorizationProcessorHelper,
+)
+
+
+class Process:
```
ross-p-smith, line 8: This is a factory class so maybe rename this to EmbedderFactory and the method should probably be "create" as it doesn't get one, it creates one each time. "Process" for me is a low level concept and doesn't describe what its creating

code/backend/batch/utilities/helpers/Processor.py
ross-p-smith: Agree that it needed to come out of the DocumentProcessorHelper. This is a type that represents a structure from the config class - correct? Maybe we should group these classes in a folder called config. Open to peoples thoughts here

code/backend/batch/utilities/helpers/processors/ProcessorBase.py
```diff
@@ -0,0 +1,7 @@
+from abc import ABC, abstractmethod
+
+
+class ProcessorBase(ABC):
```
ross-p-smith, line 4: Again - I think we could come up with a better name for this (EmbedderBase?!?!)

ross-p-smith (changes requested):
This is great and realy cleans up some of the function code and allows us to test them better. I feel some of the naming could change and I appreciate you have inherited most of those names!!

Happy with all the code and the intent. Naming is hard!

ross-p-smith (approved):

ross-p-smith (changes requested):
This is great and really simplfies the function. Makes perfect sense to me

Naming is hard and I appreciate you are working with what was there. However I think we can describe those classes better and get rid of "Helper/Process/Handler"

code/backend/batch/utilities/helpers/processors/DocumentProcessorHelper.py
```diff
@@ -0,0 +1,56 @@
...
+            for x in ConfigHelper.get_active_config_or_default().document_processors
+            if x.document_type.lower() == file_extension.lower()
+        ]
+        self.process(source_url=source_url, processors=processors)
+        if file_extension != "url":
+            self.blob_client.upsert_blob_metadata(
+                file_name, {"embeddings_added": "true"}
+            )
+
+    def process(self, source_url: str, processors: List[Processor]):
```
superhindupur, line 35: Will this method ever be used outside of this class/file? If not, can we make it private?

superhindupur: I realized this is used for url embeddings, so we can leave it as is.

code/backend/batch/utilities/helpers/processors/DocumentProcessorHelper.py
gaurarpit: seems like we're taking a document and pushing into VectorStore.
`VectorPusher` ?

code/backend/batch/utilities/helpers/processors/IntegratedVectorizationProcessorHelper.py
```diff
@@ -0,0 +1,40 @@
...
+from ...integrated_vectorization.AzureSearchIndexer import AzureSearchIndexer
+from ...integrated_vectorization.AzureSearchDatasource import AzureSearchDatasource
+from ...integrated_vectorization.AzureSearchSkillset import AzureSearchSkillset
+from ..ConfigHelper import ConfigHelper
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class IntegratedVectorizationProcessorHelper(ProcessorBase):
```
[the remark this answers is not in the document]
gaurarpit: +1.
Having a file inside the folder called `helpers` implies it is a helper. Putting it in name too, adds just more words.

code/backend/batch/utilities/helpers/processors/DocumentProcessorHelper.py
```diff
@@ -0,0 +1,56 @@
...
+    def __init__(self, blob_client: AzureBlobStorageClient):
+        self.blob_client = blob_client
+
+    def process_file(self, source_url: str, file_name: str):
+        file_extension = file_name.split(".")[-1]
+        processors = [
+            x
+            for x in ConfigHelper.get_active_config_or_default().document_processors
+            if x.document_type.lower() == file_extension.lower()
+        ]
```
gaurarpit, line 28: thinking aloud here::
1. it seems whenever we have to process a file, we call this fxn
2. we load the configs from the helper or get default
3. iterate over them each time and get a matching one
4. then process.

Can we rather:
1. on init  of this class, do the step(3) above once and store in a dictionary/map/loop-up
2. So, whenever this fxn is called, we dont do iteration each time, but,just do a constant O(1) lookup in the dictionary based on the extension ?

gaurarpit (commented):
some comments.

code/backend/batch/utilities/integrated_vectorization/AzureSearchSkillset.py
```diff
@@ -37,8 +38,8 @@ def create_skillset(self):
             description="Split skill to chunk documents",
             text_split_mode="pages",
             context="/document",
-            maximum_page_length=self.env_helper.AZURE_SEARCH_IV_MAX_PAGE_LENGTH,
-            page_overlap_length=self.env_helper.AZURE_SEARCH_IV_PAGE_OVERLAP_LENGTH,
+            maximum_page_length=self.config.integrated_vectorization_config.max_page_length,
```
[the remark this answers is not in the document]
komalg1: It will not be null if it reaches AzureSearchSkillset.py. Setting these values in default.json

code/backend/batch/utilities/helpers/processors/Process.py
komalg1: Updated

code/backend/batch/utilities/helpers/processors/DocumentProcessorHelper.py
komalg1: 👍 Updated

code/backend/batch/utilities/helpers/embedders/IntegratedVectorizationEmbedder.py
```diff
@@ -0,0 +1,40 @@
...
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class IntegratedVectorizationEmbedder(EmbedderBase):
+    def __init__(self, env_helper: EnvHelper):
+        self.env_helper = env_helper
+        self.llm_helper: LLMHelper = LLMHelper()
+        pass
```
cecheta, line 18:
```suggestion
```

code/backend/batch/utilities/helpers/config/default.json
```diff
@@ -107,6 +107,10 @@
       }
     }
   ],
+  "integrated_vectorization_config":{
```
cecheta, line 110:
```suggestion
  "integrated_vectorization_config": {
```

code/backend/batch/utilities/integrated_vectorization/AzureSearchSkillset.py
```diff
@@ -37,8 +39,8 @@ def create_skillset(self):
             description="Split skill to chunk documents",
             text_split_mode="pages",
             context="/document",
-            maximum_page_length=self.env_helper.AZURE_SEARCH_IV_MAX_PAGE_LENGTH,
```
cecheta, line 40: Can we remove all references to these env vars?

code/backend/batch/utilities/helpers/embedders/IntegratedVectorizationEmbedder.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def embed_file(self, source_url: str, file_name: str):
+        self.process_using_integrated_vectorization(source_url=source_url)
```
cecheta, line 21: Is there a reason why there's a separate function for `process_using_integrated_vectorization`? Could it not all happen in `embed_file`?

code/backend/batch/utilities/helpers/embedders/PushEmbedder.py
```diff
@@ -0,0 +1,55 @@
...
+from ...common.SourceDocument import SourceDocument
+
+logger = logging.getLogger(__name__)
+
+
+class PushEmbedder(EmbedderBase):
+    def __init__(self, blob_client: AzureBlobStorageClient):
+        self.blob_client = blob_client
+        config = ConfigHelper.get_active_config_or_default()
+        self.processor_map = {}
```
cecheta, line 22: Would we ever expect there to be two processors for the same file extension? Because the code supported that previously but now we're losing it.

We could still have it by having a dict where the key is the extension and the value is a list of processors. Could maybe use `defaultdict`

code/backend/batch/utilities/integrated_vectorization/AzureSearchSkillset.py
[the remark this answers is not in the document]
ross-p-smith: Does that mean it would be better to pass IntegratedVectorizationConfig into the constructor rather than the whole config object?

code/backend/batch/utilities/helpers/embedders/PushEmbedder.py
```diff
@@ -0,0 +1,55 @@
...
+        for processor in config.document_processors:
+            ext = processor.document_type.lower()
+            self.processor_map[ext] = processor
+
+    def embed_file(self, source_url: str, file_name: str):
+        file_extension = file_name.split(".")[-1]
+        processor = self.processor_map.get(file_extension)
+        self.__embed(source_url=source_url, processor=processor)
+        if file_extension != "url":
```
cecheta, line 31: Instead of checking for `"url"` here, which seems somewhat arbitrary and hardcoded, what do you think about adding a third, optional parameter `metadata`, and then passing in `{"embeddings_added": "true"}` where required?

code/backend/batch/AddURLEmbeddings.py
```diff
@@ -28,12 +27,8 @@ def add_url_embeddings(req: func.HttpRequest) -> func.HttpResponse:
...
     if url:
         try:
-            config = ConfigHelper.get_active_config_or_default()
-            document_processor = DocumentProcessor()
-            processors = list(
-                filter(lambda x: x.document_type == "url", config.document_processors)
-            )
-            document_processor.process(source_url=url, processors=processors)
+            processor_handler = EmbedderFactory.create(env_helper)
+            processor_handler.embed_file(url, ".url")
```
ross-p-smith, line 31: nit:
```suggestion
            embedder = EmbedderFactory.create(env_helper)
            embedder.embed_file(url, ".url")
```

code/backend/batch/BatchPushResults.py
```diff
@@ -39,24 +37,11 @@ def do_batch_push_results(msg: func.QueueMessage) -> None:
...
-        processors = list(
-            filter(
-                lambda x: x.document_type.lower() == file_extension.lower(),
-                ConfigHelper.get_active_config_or_default().document_processors,
-            )
-        )
-        document_processor.process(source_url=file_sas, processors=processors)
-        blob_client.upsert_blob_metadata(file_name, {"embeddings_added": "true"})
+    embedder_factory = EmbedderFactory.create(env_helper)
+    embedder_factory.embed_file(file_sas, file_name)
```
ross-p-smith, line 47: nit:
```suggestion
    embedder = EmbedderFactory.create(env_helper)
    embedder.embed_file(file_sas, file_name)
```

(The create creates an object - it doesnt retrun the factory)

code/backend/pages/04_Configuration.py
```diff
@@ -65,6 +64,22 @@
...
+    )
+
+if (
+    env_helper.AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION
+    and "page_overlap_length" not in st.session_state
+):
+    st.session_state["page_overlap_length"] = (
+        config.integrated_vectorization_config.page_overlap_length
+    )
+
```
cecheta, line 82: Could this be nested?

```suggestion
if env_helper.AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION:
    if "max_page_length" not in st.session_state:
        st.session_state["max_page_length"] = (
            config.integrated_vectorization_config.max_page_length
        )

    if "page_overlap_length" not in st.session_state:
        st.session_state["page_overlap_length"] = (
            config.integrated_vectorization_config.page_overlap_length
        )

```

ross-p-smith (approved):
Few small nits - but thank you ever so much for this!

code/backend/batch/utilities/helpers/embedders/IntegratedVectorizationEmbedder.py
komalg1: `embed_file` is the abstract method for PushEmbedder & IVEmbedder. `embed_file`in PushEmbedder has specific functionality, so to maintain the same pattern for both Push & IV, created a separate method

code/backend/batch/utilities/helpers/embedders/PushEmbedder.py
komalg1: I thought about that and I don't see that use case coming up. What will be the distinction for the same extension to have 2 different processors?

code/backend/batch/utilities/integrated_vectorization/AzureSearchSkillset.py
```diff
@@ -19,7 +20,7 @@
 
 
 class AzureSearchSkillset:
-    def __init__(self, env_helper: EnvHelper):
+    def __init__(self, env_helper: EnvHelper, integrated_vectorization_config: Config):
```
ross-p-smith, line 23: Isn't this a different type now IntegrationVectorizationConfig?

code/backend/batch/utilities/helpers/embedders/PushEmbedder.py
komalg1: I agree but `embed_file` is an abstract method with 2 parameters being passed. `file_name` is not being used in IV Embedder. If we add another parameter only for Push, I think it starts loosing abstract method concept. What do you think?

code/backend/batch/utilities/integrated_vectorization/AzureSearchSkillset.py
komalg1: Missed checking in the change. Updated now

ross-p-smith (dismissed):
Thank you fixing all my comments!

code/backend/batch/utilities/helpers/embedders/PushEmbedder.py
cecheta: I don't see a need for it myself

code/backend/batch/utilities/helpers/embedders/PushEmbedder.py
cecheta: OK nw

cecheta (approved):
Thanks!

ross-p-smith (approved):

code/backend/Admin.py
```diff
@@ -2,12 +2,10 @@
...
 import logging
 import sys
-from dotenv import load_dotenv
 from azure.monitor.opentelemetry import configure_azure_monitor
+from batch.utilities.helpers.EnvHelper import EnvHelper
 
-
-load_dotenv()
-
+env_helper: EnvHelper = EnvHelper()
```
cecheta, line 8: Why was it added, it doesn't look like it's used anywhere?

Also, we should import and use the env helper after we've set up application insights below, otherwise it doesn't work correctly

gaurarpit (dismissed):
LGTM

code/backend/Admin.py
```diff
@@ -5,7 +5,6 @@
 from azure.monitor.opentelemetry import configure_azure_monitor
 from batch.utilities.helpers.EnvHelper import EnvHelper
```
cecheta, line 6: The import needs to be below app insights as well

code/backend/Admin.py
```diff
@@ -21,6 +18,7 @@
 
 logger = logging.getLogger(__name__)
 logger.debug("Starting admin app")
+env_helper: EnvHelper = EnvHelper()
```
cecheta, line 21: I would recommend adding a comment here about why this is here, otherwise it looks like an unused variable and could be removed in the future

komalg1: I was planning to do another round of testing without this. But did not want to block this PR just because of this change.

komalg1: Removed the unused variable. Looks like it was because of the type declaration issue

cecheta (approved):
Nice

superhindupur (approved):

PR #789, opened 2024-04-30
Simplify BatchStartProcessing function method.

superhindupur:
## Purpose
The BatchStartProcessing function API was only ever being called with process_all=true, so I simplified it. This reduced further code from the Ingest page in the Admin app. 

## Does this introduce a breaking change?

```
[ ] Yes
[x] No
```

## Pull Request Type
What kind of change does this Pull Request introduce?

```
[ ] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[x] Refactoring (no functional changes, no api changes)
[ ] Documentation content changes
[ ] Other... Please describe:
```

## How to Test
Deploy the admin web, the function app and test whether "Reprocess all documents..." button works.

PR #799, opened 2024-05-01
refactor: Simplify url embeddings logic in Ingest admin app page

superhindupur:
## Purpose
This PR removes more logic out of the Ingest admin app page and into the backend function, adding unit tests for the newly added logic in the function.

It also adds back the changes from PR #789 that were mistakenly reverted in a previous PR. 

```
[ ] Yes
[x ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
Deploy the admin app and the function.
Test ingesting URLs with and without integrated vectorization.


code/backend/batch/AddURLEmbeddings.py
```diff
@@ -17,34 +22,54 @@ def add_url_embeddings(req: func.HttpRequest) -> func.HttpResponse:
...
 
     # Get Url from request
-    url = req.params.get("url")
-    if not url:
-        try:
-            req_body = req.get_json()
-        except ValueError:
-            pass
-        else:
-            url = req_body.get("url")
```
superhindupur, line 27: I removed the above conditional logic since the one place calling this API sends the URL in the body

ross-p-smith:
This will conflict with #798 - any chance we can work on getting that one in first?

code/backend/batch/AddURLEmbeddings.py
```diff
@@ -17,34 +22,54 @@ def add_url_embeddings(req: func.HttpRequest) -> func.HttpResponse:
...
-            return func.HttpResponse(
-                f"Error: {traceback.format_exc()}", status_code=500
-            )
+    url = None
+    try:
+        req_body = req.get_json()
+    except ValueError:
+        pass
+    else:
+        url = req_body.get("url")
```
gaurarpit, line 31: can we write this as a 1 liner, saves from 3 lines of code-voodoo with exception block.

```suggestion
url = req.get_json().get("url") if req.get_json() else None 
```

gaurarpit (commented):
a few comments.

code/backend/batch/AddURLEmbeddings.py
cecheta: This could still throw an exception

cecheta, line 31: Does this work? try/except/else is so confusing

```suggestion
    try:
        url = req.get_json().get("url")
    except Exception:
        url = None
```

We should also catch more than `ValueError` in case it is valid JSON but not a dict

code/backend/batch/AddURLEmbeddings.py
```diff
@@ -17,34 +22,54 @@ def add_url_embeddings(req: func.HttpRequest) -> func.HttpResponse:
...
+        with io.BytesIO(parsed_data.get_text().encode("utf-8")) as stream:
+            blob_client = AzureBlobStorageClient()
+            blob_client.upload_file(stream, url, metadata={"title": url})
+        return func.HttpResponse(f"Url {url} added to knowledge base", status_code=200)
+
+    except Exception:
         return func.HttpResponse(
-            "Please pass a url on the query string or in the request body",
-            status_code=400,
+            f"Error: {traceback.format_exc()}. Exception occurred while adding {url} to the knowledge base.",
```
cecheta, line 73: I think we wanted to stop returning exceptions to clients

cecheta (commented):
Looking good, a few comments

code/backend/batch/AddURLEmbeddings.py
```diff
@@ -17,34 +22,54 @@ def add_url_embeddings(req: func.HttpRequest) -> func.HttpResponse:
...
+            status_code=400,
         )
 
+    env_helper: EnvHelper = EnvHelper()
+    if env_helper.AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION:
+        return download_url_and_upload_to_blob(url)
     else:
+        return process_url_contents_directly(url)
+
+
```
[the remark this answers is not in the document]
superhindupur: This is actually being added by the autoformatter - and is consistent with the rest of the files.

cecheta (approved):
Nice

ross-p-smith (approved):
Whoop!

PR #826, opened 2024-05-07
Update dependencies - 7thMay

ross-p-smith:
## Purpose
Related to #417 

Updated the dependencies, uploaded a document and then asked a question on it.

image

This will clear all the outstanding dependabot issues - but will be merged after #798


ross-p-smith:
> LGTM.
> 
> out of curiosity, can dependabot update these dependencies too ?

It could; but until we have test coverage of everything, I like to check these manually

PR #848, opened 2024-05-08
docs: Update docs if deploying through az

ross-p-smith:
Update the documentation to correct the `az deployment` command which was outdated.


gaurarpit (approved):
Ahh, much cleaner now. Thank you for this.

PR #849, opened 2024-05-08
fix: [IV] Appropriate messages on Explore & Delete page when index not created

### LINKED ISSUE #836: [IV] Display appropriate messages on Explore & delete pages when no index created
Relates to #321

komalg1:
## Purpose

* Resolves #836 
* This PR checks if the index is created before making any search calls to the index.
* Updating & adding corresponding unit tests

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix (`fix`)
[ ] Feature (`feat`)
[ ] Code style update (formatting, local variables) (`style`)
[ ] Refactoring (no functional changes, no api changes) (`refactor`)
[ ] Documentation content changes (`docs`)
[ ] Other... Please describe: (`perf`, `test`, `build`, `ci`, `chore`, `revert`)
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

## What to Check
* On a fresh deployment, when AZURE_SEARCH_USE_INTEGRATED_VECTORIZATION is true any user directly landing on the Explore page should see an empty grid with no chunks
image

* Similarly on the Delete page, user sees the message `No files to delete`
image


gaurarpit (approved):

code/backend/pages/02_Explore_Data.py
```diff
@@ -38,13 +38,21 @@
 try:
     search_handler = Search.get_search_handler(env_helper)
 
-    results = search_handler.search_client.search("*", facets=["title"])
-    unique_files = [filename["value"] for filename in results.get_facets()["title"]]
+    results = (
```
ross-p-smith, line 41: we have a `search_handler`, yet we expose the `search_client` and let people call search. I feel we have a leaky abstraction here that could be tidied up. Will also keep the streamlit code simple. Happy for it to be a seperate PR if you want to create one (if others agree!)

code/backend/pages/03_Delete_Data.py
```diff
@@ -40,8 +40,8 @@
     search_handler = Search.get_search_handler(env_helper)
     search_client = search_handler.search_client
 
-    results = search_handler.get_files()
-    if results.get_count() == 0:
+    results = search_handler.get_files() if search_client is not None else None
```
ross-p-smith, line 43: Again - could this be pushed down to the search_handler.get_file_count() which will make this even simpler

code/backend/pages/02_Explore_Data.py
superhindupur: +1. Also this if-else will not be needed here if we stop exposing the search_client directly and handle it from the search_handler directly.

code/backend/pages/02_Explore_Data.py
```diff
@@ -38,13 +38,21 @@
...
+        [filename["value"] for filename in results.get_facets()["title"]]
+        if results
+        else []
+    )
     filename = st.selectbox("Select your file:", unique_files)
     st.write("Showing chunks for:", filename)
 
     results = search_handler.perform_search(filename)
-    data = search_handler.process_results(results)
+    data = search_handler.process_results(results) if results else []
```
superhindupur, line 55: Can this be handled inside search_handler.process_results?

code/backend/batch/utilities/search/Search.py
```diff
@@ -19,7 +19,11 @@ def get_search_handler(env_helper: EnvHelper):
     def get_source_documents(search_handler, question):
         if isinstance(search_handler, IntegratedVectorizationSearchHandler):
             search_results = search_handler.query_search(question)
-            return Search.generate_source_documents(search_results)
+            return (
+                Search.generate_source_documents(search_results)
+                if search_results
+                else []
```
superhindupur, line 25: Are we missing a test case for this "else" condition?

code/backend/batch/utilities/search/IntegratedVectorizationSearchHandler.py
cecheta, line 1: I feel like it shouldn't be the responsibility of this Handler to check if the index exists at the start of every single method, but it should be the responsibility of whoever is using the Handler to check if an index exists before trying to interact with it, and/or handle exceptions accordingly

Not sure how others feel

code/backend/batch/utilities/search/Search.py
komalg1: Added

code/backend/pages/02_Explore_Data.py
komalg1: Updated

code/backend/pages/03_Delete_Data.py
komalg1: the files from `get_files()` is being used in the next line `files = search_handler.output_results(results)`
I have removed the search_client check as it being checked in the `get_files` method

code/backend/pages/02_Explore_Data.py
komalg1: Created separate method for this and removed the exposed search_client from Explore_Data page

code/backend/batch/utilities/search/SearchHandlerBase.py
```diff
@@ -1,13 +1,24 @@
...
 from ..common.SourceDocument import SourceDocument
 
 
 class SearchHandlerBase(ABC):
-    def __init__(self, env_helper):
+    def __init__(self, env_helper: EnvHelper):
         self.env_helper = env_helper
         self.search_client = self.create_search_client()
 
+    def search_with_facets(self, query, facets):
```
ross-p-smith, line 12: can we put some types on these as it looks like it is a dictionary?

ross-p-smith (dismissed):
Small nit on types, happy to approve after again

code/backend/batch/utilities/search/SearchHandlerBase.py
komalg1: Updated

code/backend/batch/utilities/search/IntegratedVectorizationSearchHandler.py
komalg1: Hmm I think if the handler is not responsible for checking the index, it will create a lot of duplicate code (if/else) of checking the index existence before each search interaction. But happy to discuss if anyone has other thoughts

cecheta: No worries, let's not make it blocking

ross-p-smith (approved):
yep

PR #850, opened 2024-05-08
chore: move imports to top of files

liammoat:
## Purpose

* Update a few occurrences where imports were [outside top level](https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/import-outside-toplevel.html). 

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix (`fix`)
[ ] Feature (`feat`)
[x] Code style update (formatting, local variables) (`style`)
[ ] Refactoring (no functional changes, no api changes) (`refactor`)
[ ] Documentation content changes (`docs`)
[ ] Other... Please describe: (`perf`, `test`, `build`, `ci`, `chore`, `revert`)
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Run the test suite

```
poetry run pytest -m "not azure"
```

## What to Check
* Tests run successful and import modules as expected

## Other Information
This change also introduced some circular dependencies in the strategy files. Some symbols were moved to their own files to remove this coupling.

### DIFF
--- a/code/backend/batch/utilities/document_chunking/ChunkingStrategy.py
+++ b/code/backend/batch/utilities/document_chunking/ChunkingStrategy.py
@@ -1,0 +1,25 @@
+from enum import Enum
+
+
+class ChunkingStrategy(Enum):
+    LAYOUT = "layout"
+    PAGE = "page"
+    FIXED_SIZE_OVERLAP = "fixed_size_overlap"
+    PARAGRAPH = "paragraph"
+
+
+class ChunkingSettings:
+    def __init__(self, chunking: dict):
+        self.chunking_strategy = ChunkingStrategy(chunking["strategy"])
+        self.chunk_size = chunking["size"]
+        self.chunk_overlap = chunking["overlap"]
+
+    def __eq__(self, other: object) -> bool:
+        if isinstance(self, other.__class__):
+            return (
+                self.chunking_strategy == other.chunking_strategy
+                and self.chunk_size == other.chunk_size
+                and self.chunk_overlap == other.chunk_overlap
+            )
+        else:
+            return False
--- a/code/backend/batch/utilities/document_loading/Strategies.py
+++ b/code/backend/batch/utilities/document_loading/Strategies.py
@@ -1,4 +1,8 @@
 from enum import Enum
+from .Layout import LayoutDocumentLoading
+from .Read import ReadDocumentLoading
+from .Web import WebDocumentLoading
+from .WordDocument import WordDocumentLoading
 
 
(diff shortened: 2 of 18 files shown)


code/backend/batch/utilities/orchestrator/OrchestrationStrategy.py
```diff
@@ -0,0 +1,7 @@
+from enum import Enum
+
+
+class OrchestrationStrategy(Enum):
```
gaurarpit, line 4: this is nice. :)

code/backend/batch/utilities/document_chunking/ChunkingStrategy.py
```diff
@@ -0,0 +1,25 @@
+from enum import Enum
+
+
+class ChunkingStrategy(Enum):
```
gaurarpit, line 4: this is really neat.

code/backend/batch/utilities/orchestrator/__init__.py
```diff
@@ -2,7 +2,7 @@
 from typing import List
 import os.path
 import pkgutil
-from .Strategies import OrchestrationStrategy
+from .OrchestrationStrategy import OrchestrationStrategy
 
 
 class OrchestrationSettings:
```
gaurarpit, line 8: minor/low-priority: I would also suggest to move the OrchestrationSettings class into OrchestrationSettings file ( snake_case_named or whatever for now) and out of init.py as it's really hidden inside and impacts readability.

gaurarpit (approved):
looks much cleaner. LGTM.
one minor suggestion, but not a blocker.

code/backend/batch/utilities/document_loading/Strategies.py
```diff
@@ -10,20 +14,12 @@ class LoadingStrategy(Enum):
 
 def get_document_loader(loader_strategy: str):
```
ross-p-smith, line 15: I feel that this should really be a static `DocumentLoadingFactory.create(loader_strategy: str)` - but feel might be pushing it for this PR - (feel free to resolve!)

ross-p-smith (approved):
Many thanks!

code/backend/batch/utilities/orchestrator/__init__.py
liammoat: Thanks Arpit. I agree with your comment but, on this occasion, I am going to leave this as-is - in a hope to keep the diff on this PR minimal.

PR #851, opened 2024-05-08
build: Boolean only accepting a false is causing bicep ERROR in CI

ross-p-smith:
## Purpose
Fix the broken CI. According to the logs, we have this error. I believe it is because we define it as a bool, but we only allow "false" as a value.

```
  /workspaces/chat-with-your-data-solution-accelerator/infra/main.bicepparam(29,47) : Error BCP033: Expected a value of type "false" but the provided value is of type "bool".
```

I am however unsure why this passed the previous build!


adamdougal (approved):
Makes sense!

Issue #796, opened 2024-05-01
Add all params to `main.bicepparam`


cecheta:
# Motivation

It should be possible to set every parameter in the `main.bicep` file with a corresponding environment variable in the `main.bicepparam` file, so users do not need to manually edit `main.bicep` when they are deploying.

# Requirements

A list of requirements to consider this feature delivered
- Every param in `main.bicep` should have a corresponding param in `main.bicepparam`
- The defaults should be the same in both
- Remove the backwards compatibility check for the search service (Lines 29- 44)

# Tasks

_To be filled in by the engineer picking up the issue_

PR #868, opened 2024-05-09
build: Adding params that we have ENV for

ross-p-smith:
## Purpose
Relates to #796 

This is a part fix for #796 as we need to design this so that it works for people who use azd, plain bicep and the "Deploy to Azure" button. Not all environment variables have a corresponding input bicep param as they are outputs of the bicep. I'm also unsure how brittle we will make this if we allow people change the name of Azure resources through the environment variables, I think it was better if they were bicepparam's and the ENV values were the output. I will look into this next.

This PR just adds those values that are ENV and params so that you can deploy using `azd up` and then update using `azd env set KEY VALUE` and then `azd up` again and have the deployment respect the value you have just set.

This also has a small tweak to the B3 -> B1 issue

### DIFF
--- a/infra/main.bicep
+++ b/infra/main.bicep
@@ -458,7 +458,7 @@ module hostingplan './core/host/appserviceplan.bicep' = {
       tier: skuTier
     }
     reserved: true
-    tags: { Automation: 'Ignore' }
+    tags: { CostControl: 'Ignore' }
   }
 }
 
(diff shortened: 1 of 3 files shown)


adamdougal (approved):
lgtm

cecheta (approved):

gaurarpit:
Should we create a User Story/task for this ?
>  I'm also unsure how brittle we will make this if we allow people change the name of Azure resources through the environment variables, I think it was better if they were bicepparam's and the ENV values were the output. I will look into this next.

superhindupur (approved):

PR #872, opened 2024-05-10
fix: Removing index creation from init of AzureSearchHelper.py

komalg1:
## Purpose

* This PR fixes a bug: First message in web chat creates an index for Push approach although IV is enabled while creating the conversation logger index.(Index creation called in init) 

- [ ] Yes
- [x] No

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

* When IV is enabled, web chat message should not create the index.


code/tests/utilities/helpers/test_AzureSearchHelper.py
```diff
@@ -134,116 +116,6 @@ def test_returns_search_client(
     assert search_client is search_client_mock.return_value
 
 
-@patch("backend.batch.utilities.helpers.AzureSearchHelper.SearchClient")
-@patch("backend.batch.utilities.helpers.AzureSearchHelper.SearchIndexClient")
-def test_creates_search_index_if_not_exists(
```
adamdougal, line 139: I think we want to keep this test, but instead of the `# when` section just calling the constructor, it also now needs to call `get_client`

code/tests/utilities/helpers/test_AzureSearchHelper.py
```diff
@@ -262,24 +134,6 @@ def test_does_not_create_search_index_if_it_exists(
     search_index_client_mock.return_value.create_index.assert_not_called()
 
 
-@patch("backend.batch.utilities.helpers.AzureSearchHelper.SearchClient")
-@patch("backend.batch.utilities.helpers.AzureSearchHelper.SearchIndexClient")
-def test_propogates_exceptions_when_creating_search_index(
```
adamdougal, line 267: Same for this test.

adamdougal (changes requested):

code/tests/functional/backend_api/tests/default/test_conversation_custom.py
```diff
@@ -620,35 +620,3 @@ def test_post_makes_correct_call_to_store_conversation_in_search(
             times=2,
         ),
     )
-
-
-def test_post_returns_error_when_downstream_fails(
```
komalg1, line 625: @adamdougal Is it ok if we remove this test as there is no call happening directly to index creation?

code/tests/functional/backend_api/tests/sk_orchestrator/test_response_without_tool_call.py
```diff
@@ -251,35 +251,3 @@ def test_post_makes_correct_call_to_store_conversation_in_search(
             times=2,
         ),
     )
-
-
-def test_post_returns_error_when_downstream_fails(
-    app_url: str, app_config: AppConfig, httpserver: HTTPServer
```
komalg1, line 257: @adamdougal same here

code/tests/functional/backend_api/tests/default/test_conversation_custom.py
adamdougal: Ah we still want to test that an error is returned when a downstream fails. We will need to switch this to the first call we're now making instead.

superhindupur:
@komalg1 have you also tested this works when IV is disabled?

adamdougal (approved):
LGTM!

komalg1:
> test_post_returns_error_when_downstream_fails

Yes tested 👍

superhindupur (approved):

PR #877, opened 2024-05-13
refactor: use semantic kernel instead of langchain

### LINKED ISSUE #759: Update Semantic Kernel functions to not rely on LangChain
Required by #320 

# Description

Replace the existing functions, which use LangChain, to use Semantic Kernel functions that do not rely on LangChain.

# Tasks

_To be filled in by the engineer picking up the subtask

- [ ] Update `QuestionAnswerTool`
- [ ] Update `TextProcessingTool`
- [ ] Update `PostPromptTool`

gaurarpit:
## Purpose

* Remove TODOs
* Resolves #759 759

```
discussed with @cecheta and @superhindupur and agreed on below:

Langchain is being used in QuestionAnswerTool, TextProcessingTool, PostPromptTool as a library rather than a orchestration provider.
therefore there is no need to update these files
Semantic Kernel orchestration processes the entire request flow using its own kernel's functions already.
verified by doing a debug call on the entire chat flow.
```

Ref: https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/issues/759#issuecomment-2107126387

- [ ] Yes
- [x] No


infra/main.bicep
```diff
@@ -119,7 +119,7 @@ param azureOpenAIVisionModelVersion string = 'vision-preview'
 @description('Azure OpenAI Vision Model Capacity - See here for more info  https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/quota')
 param azureOpenAIVisionModelCapacity int = 10
 
-@description('Orchestration strategy: openai_function or langchain str. If you use a old version of turbo (0301), plese select langchain')
+@description('Orchestration strategy: openai_function or langchain str. If you use a old version of turbo (0301), please select langchain')
 @allowed([
   'openai_function'
```
cecheta, line 124: There should be `semantic_kernel` here as well 🙈

cecheta: And in the description

PR #891, opened 2024-05-13
feat: Implement hybrid search on BYOD endpoint.

### LINKED ISSUE #295: Allow for different queryType settings for chat on your data to Azure AI Search

superhindupur:
## Purpose
Before this change, the `/api/conversation/byod` endpoint did not use hybrid search - it only used "simple" or "semantic" query types, based on whether semantic search was enabled. 

This PR changes the query types to `vector_simple_hybrid` and `vector_simple_hybrid` respectively. 

Fixes #295

- [ ] Yes
- [x] No

## How to Test
Deploy CWYD.
Ingest a file and test whether chat works with and without semantic search enabled.


ross-p-smith:
Is this for #295?

gaurarpit:
> Is this for #295?

Yes.

superhindupur:
> Is this for #295?

Yes @ross-p-smith. I updated the PR description.

code/create_app.py
```diff
@@ -130,10 +130,14 @@ def conversation_with_data(request: Request, env_helper: EnvHelper):
                         "filter": env_helper.AZURE_SEARCH_FILTER,
                         "in_scope": env_helper.AZURE_SEARCH_ENABLE_IN_DOMAIN,
                         "top_n_documents": env_helper.AZURE_SEARCH_TOP_K,
+                        "embedding_dependency": {
+                            "type": "deployment_name",
+                            "deployment_name": env_helper.AZURE_OPENAI_EMBEDDING_MODEL,
+                        },
```
gaurarpit, line 136: neat. 👍

code/backend/batch/utilities/search/AzureSearchHandler.py
```diff
@@ -101,6 +101,7 @@ def _hybrid_search(self, question: str, tokenised_question: list[int]):
                     fields=self._VECTOR_FIELD,
                 )
             ],
+            query_type="simple",  # this is the default value
```
adamdougal, line 104: Is this comment required? Is the query type required if it's already the default?

adamdougal (approved):
One small nit, but LGTM!

code/create_app.py
```diff
@@ -130,10 +130,14 @@ def conversation_with_data(request: Request, env_helper: EnvHelper):
...
                         "query_type": (
-                            "semantic"
+                            "vector_semantic_hybrid"
```
cecheta, line 138: Did you consider adding a new env var to allow the user to pick any of the five different search types?

ross-p-smith: I like this!

code/backend/batch/utilities/search/AzureSearchHandler.py
adamdougal: I probably wouldn't keep the commit. But we can remove it later!

code/create_app.py
superhindupur: This is a good idea, this will give the user full and transparent control over the query type.

However, so far, Chat with your data has been opinionated (at least via the /api/conversations/custom endpoint) about always using hybrid search (as opposed to text-based-only or vector-only). As of this PR, the /byod endpoint has parity with the /custom endpoint - it always uses hybrid search, and enables semantic search when the user requests it (while still maintaining hybrid search).

I have created #897 to cover the work to allow for all five types of queries to be available as options to the user to choose from.

PR #892, opened 2024-05-13
feat: Generate embeddings for images

cecheta:
Required by #748 

## Purpose

* This PR fetches embeddings from Computer Vision when an image is uploaded and advanced image processing is enabled
* Adds functional and unit tests
  * Note that the unit tests rely on pytest-httpserver
  * To enable this, the fixtures for enabling HTTPS for the mock server have been moved to the top-level `conftest.py`, so they are in the scope of all tests (unit + functional)
* Created client to call computer vision to generate embeddings, as this is not supported yet by the official SDK
* Add config validation before saving config file, so that advanced image processing cannot be enabled for non-image types (e.g. .txt)
* Added computer vision key/endpoint to env vars

- [ ] Yes
- [x] No

Testing:
- Deployed to Azure in keys + rbac
- Uploaded images, verified embeddings in logs


code/backend/batch/utilities/helpers/azure_computer_vision_client.py
```diff
@@ -0,0 +1,75 @@
...
+from requests import Response
+
+from .env_helper import EnvHelper
+
+logger = logging.getLogger(__name__)
+
+
+class AzureComputerVisionClient:
+
+    __TOKEN_SCOPE = "https://cognitiveservices.azure.com/.default"
```
gaurarpit, line 16: Should these configs be in ENV ?

code/backend/batch/utilities/helpers/azure_computer_vision_client.py
```diff
@@ -0,0 +1,75 @@
...
+    __RESPONSE_VECTOR_KEY = "vector"
+
+    def __init__(self, env_helper: EnvHelper) -> None:
+        self.computer_vision_host = env_helper.AZURE_COMPUTER_VISION_ENDPOINT
+        self.computer_vision_key = env_helper.AZURE_COMPUTER_VISION_KEY
+        self.use_keys = env_helper.is_auth_type_keys()
+
+    def vectorize_image(self, image_url: str) -> List[float]:
+        logger.info("Making call to computer vision to vectorize image")
+        response = self.__make_request(image_url)
```
gaurarpit, line 29: do we want to have this call in try-catch ? what happens if the call fails ?

code/backend/batch/utilities/helpers/azure_computer_vision_client.py
```diff
@@ -0,0 +1,75 @@
...
+            },
+            json={"url": image_url},
+            headers=headers,
+        )
+
+    def __validate_response(self, response: Response):
+        if response.status_code != 200:
+            raise Exception(
+                f"Call to vectorize image failed with status: {response.status_code} body: {response.text}"
+            )
```
gaurarpit, line 59: suggestion: we can get rid of this if we did the call in a try-block ?

code/backend/batch/utilities/helpers/config/config_helper.py
```diff
@@ -180,13 +184,29 @@ def get_active_config_or_default():
...
+            document_type = document_processor.get("document_type")
+            unsupported_advanced_image_processing_file_type = (
+                document_type not in ADVANCED_IMAGE_PROCESSING_FILE_TYPES
+            )
+            if (
+                document_processor.get("use_advanced_image_processing")
+                and unsupported_advanced_image_processing_file_type
+            ):
+                raise Exception(
+                    f"Advanced image processing is enabled for document type {document_type}, but only {ADVANCED_IMAGE_PROCESSING_FILE_TYPES} files types are supported."
```
gaurarpit, line 207: sounds confusing ? if the document_type is not supported, why do we say `Advanced image processing is enabled for document type....` ? Is this a typo ?

code/backend/batch/utilities/helpers/embedders/push_embedder.py
```diff
@@ -19,30 +21,48 @@
...
+            embedding_config.use_advanced_image_processing
+            and file_extension
+            in self.config.get_advanced_image_processing_image_types()
+        ):
+            logger.warning("Advanced image processing is not supported yet")
+            image_vectors = self.azure_computer_vision_client.vectorize_image(
+                source_url
+            )
+            logger.info("Image vectors: " + str(image_vectors))
+            # Coming soon, storing the image embeddings in Azure Search
```
gaurarpit, line 64: a todo ?

gaurarpit (commented):
Looks great.. some comments

code/backend/batch/utilities/helpers/azure_computer_vision_client.py
adamdougal: We did discuss if we think this should be configurable, however decided to leave them here as they are unlikely to change, with perhaps the exception being the versions. Happy to pull the versions out to the env though.

code/backend/batch/utilities/helpers/azure_computer_vision_client.py
adamdougal: I don't think so, as the requests library does not throw exceptions for bad status codes being returned.

code/backend/batch/utilities/helpers/config/config_helper.py
adamdougal: Ah it's because it's been enabled by the user, but it's not supported. It doesn't mean it's been enabled by the app. We'll reword it to try and improve.

code/backend/batch/utilities/helpers/embedders/push_embedder.py
adamdougal: Only put a comment here as I thought it would look odd just logging and doing nothing. I didn't put a TODO as we have it as a task on our issue and it's not something anyone else needs to look at.

code/backend/batch/utilities/helpers/azure_computer_vision_client.py
cecheta: Moved the two versions to the env helper

code/backend/batch/utilities/helpers/azure_computer_vision_client.py
```diff
@@ -0,0 +1,75 @@
...
+    __VECTORIZE_IMAGE_MODEL_VERSION = "2023-04-15"
+    __RESPONSE_VECTOR_KEY = "vector"
+
+    def __init__(self, env_helper: EnvHelper) -> None:
+        self.computer_vision_host = env_helper.AZURE_COMPUTER_VISION_ENDPOINT
+        self.computer_vision_key = env_helper.AZURE_COMPUTER_VISION_KEY
+        self.use_keys = env_helper.is_auth_type_keys()
+
+    def vectorize_image(self, image_url: str) -> List[float]:
+        logger.info("Making call to computer vision to vectorize image")
```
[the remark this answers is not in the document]
cecheta: Added

code/backend/batch/utilities/helpers/azure_computer_vision_client.py
cecheta: We've put the request in a try/except block, this should catch errors where we do not receive a proper HTTP response

As part of this, we've configured a timeout

code/backend/batch/utilities/helpers/config/config_helper.py
cecheta: Renamed slightly, hopefully it's clearer now

code/backend/batch/utilities/helpers/azure_computer_vision_client.py
```diff
@@ -0,0 +1,82 @@
...
+                params={
+                    "api-version": self.api_version,
+                    "model-version": self.model_version,
+                },
+                json={"url": image_url},
+                headers=headers,
+                timeout=self.timeout,
+            )
+        except Exception as e:
+            raise Exception("Call to vectorize image failed") from e
```
gaurarpit, line 60: may be add the `image_url` in the exception to know exactly which image failed ?

code/backend/batch/utilities/helpers/azure_computer_vision_client.py
gaurarpit: gotcha. makes sense

code/backend/batch/utilities/helpers/config/config_helper.py
gaurarpit: ahh, I see what you mean. :) It's always tricky to come up with a perfect message.

code/backend/batch/utilities/helpers/azure_computer_vision_client.py
```diff
@@ -0,0 +1,82 @@
...
+            raise Exception(
+                f"Call to vectorize image returned malformed response body: {response.text}",
+            ) from e
+
+    def __get_vectors(self, response_json: dict) -> List[float]:
+        if self.__RESPONSE_VECTOR_KEY in response_json:
+            return response_json[self.__RESPONSE_VECTOR_KEY]
+        else:
+            raise Exception(
+                f"Call to vectorize image returned no vector: {response_json}"
```
gaurarpit, line 81: personally, I am not a fan of throwing entire responses into Exception msgs. But, here I dont know what attribute would make sense to return in the exception. So, just calling it out if there is anything we can do to make it trimmer , if not, I guess, it's alright.

code/backend/batch/utilities/helpers/embedders/push_embedder.py
```diff
@@ -19,30 +21,48 @@
...
-    def __embed(self, source_url: str, embedding_config: EmbeddingConfig):
+    def __embed(
+        self, source_url: str, file_extension: str, embedding_config: EmbeddingConfig
+    ):
         documents_to_upload: List[SourceDocument] = []
-        if not embedding_config.use_advanced_image_processing:
+        if (
+            embedding_config.use_advanced_image_processing
+            and file_extension
+            in self.config.get_advanced_image_processing_image_types()
```
gaurarpit, line 57: for readability:

```suggestion
        if (
            embedding_config.use_advanced_image_processing
            and file_extension in self.config.get_advanced_image_processing_image_types()
```

gaurarpit (approved):
Looks great. Nothing major, some minor ones. Ship-It.

```
                              ______________________________________________
                          .-'                     _                        '.
                        .'                       |-'                        |
                      .'                         |                          |
                   _.'               p         _\_/_         p              |
                _.'                  |       .'  |  '.       |              |
           __..'                     |      /    |    \      |              |
     ___..'                         .T\    ======+======    /T.             |
  ;;;\::::                        .' | \  /      |      \  / | '.           |
  ;;;|::::                      .'   |  \/       |       \/  |   '.         |
  ;;;/::::                    .'     |   \       |        \  |     '.       |
        ''.__               .'       |    \      |         \ |       '.     |
             ''._          <_________|_____>_____|__________>|_________>    |
                 '._     (___________|___________|___________|___________)  |
                    '.    \;;;o;;;;o;;;;;o;;;;;o;;;;;o;;;;;o;;;;;o;;;;/   |
                      '.~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~   |
                        '. ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~  |
                          '-.______________________________________________.'

```

gaurarpit:
tests are failing.. looking into them

gaurarpit (dismissed):

gaurarpit:
> tests are failing.. looking into them

fixed.

superhindupur:
Regarding the use of a mock http server for unit tests:

I understand the point about not wanting to tie our mocks tightly to the underlying implementation (http requests vs sdk), but couldn't we resolve this issue by just using an interface and implementation? Mock the interface behavior in the tests, and the implementation details will remain hidden. 
Also, mocks in general tend to be tied to underlying implementation. Why the special consideration for this case? 
 
I would vote for keeping unit tests free of http servers - and reserve this type of testing for our functional tests. Consistency helps readers reason about the code base. I'd argue that in this case readability is more important than the possible rework of the test mocks if/when we switch the underlying calls from http to sdk.

code/backend/batch/utilities/helpers/azure_computer_vision_client.py
adamdougal: We could log it instead, though more inclined to keep it here. 

I would normally agree with you but this is only going to be outputted by the Azure Function logs so it won't go to an "end user" as such.

code/backend/batch/utilities/helpers/embedders/push_embedder.py
adamdougal: This was formatted by flake8. Maybe we need a larger line length...

adamdougal:
Bhavana and I caught up on a call. Decided to keep the tests as they are, but added a comment to the test class explaining the rationale.

code/backend/batch/utilities/helpers/azure_computer_vision_client.py
gaurarpit: sure, sure. it was a suggestion.

code/backend/batch/utilities/helpers/embedders/push_embedder.py
gaurarpit: haha... flake8 works in a flaky way. :)

Issue #625, opened 2024-04-08
Use Key Vault references for environment variable secrets


cecheta:
# Motivation

Currently when Key Vault is used, the `EnvHelper()` makes a number of requests to Azure Key Vault. EnvHelper is currently cached, but there can be multiple calls to key vault, causing unnecessary latency.

We can rework the keyvault functionality to use the built-in Key Vault references, so the secrets are fetched from Key Vault when the application starts, and the keys are directly injected into the environment as environment variables, instead of fetching them at runtime.

https://learn.microsoft.com/en-us/azure/app-service/app-service-key-vault-references

# How would you feel if this feature request was implemented?

![image](https://media.giphy.com/media/26tk1ytJtqSOz16es/giphy.gif?cid=ecf05e47ongvwp7ce3zi8airgh0tbtrpuxurlaga5bckjacv&ep=v1_gifs_search&rid=giphy.gif&ct=g)

# Requirements

- When `useKeyVault` is `true`, replace all secret env vars with a reference to Key Vault
- Remove `azure-keyvault-secrets` from project
- Update docs, particularly around the need to manually set the keys in an `.env` file, or with `azd env set`
  - Write a post-provision script to automate this (?)

# Tasks

_To be filled in by the engineer picking up the issue_

PR #893, opened 2024-05-13
fix: handle BlobDelete event type in `batch_push_results`

liammoat:
## Purpose

* `batch_push_results` receives, but does not handle, blob delete events.
* This PR introduces the functionality to delete the search index, when a blob is deleted from Azure Storage

- [ ] Yes
- [x] No

## How to Test
1. Run the updated Azure Function
2. `#1` Create a blob, uploading via Azure Storage
3. `#2` Delete the blob from Azure Storage
4. `#3` Trigger `BatchStartProcessing function

## What to Check
1. `#1` `BatchPushResults` is triggered, and the document is indexed.
2. `#2` `BatchPushResults` is triggered, and the associated indexes are removed.
3. `#3` `BatchPushResults` is triggered, as normal, and the documents are indexed.

## Remaining Actions
The following is outstanding for this PR:

- [ ] Tests


liammoat:
Team - Couple of things to note about this PR:

1. There is likely a better way to do this - however, I have tried to implement it with as few downstream changes as possible (i.e. I don't want to refactor everything to fix this)
2. With this change in mind, I propose the admin app should delete the blob, and let the function handle the index. 
3. I will leave this in draft until I have introduce some tests to cover the core behaviour

code/backend/batch/BatchPushResults.py
```diff
@@ -27,21 +27,38 @@ def _get_file_name_from_message(msg: func.QueueMessage) -> str:
...
+    if event_type in ("", "Microsoft.Storage.BlobCreated"):
+        _process_document_created_event(message_body)
 
+    elif event_type == "Microsoft.Storage.BlobDeleted":
+        _process_document_deleted_event(message_body)
 
-def do_batch_push_results(msg: func.QueueMessage) -> None:
+    else:
+        logger.error("Unknown event type received: %s", event_type)
+        raise NotImplementedError(f"Unknown event type received: {event_type}")
```
gaurarpit, line 45: suggestion: generally, logging an error and throwing an exception is considered an anti-pattern.
Since, throwing an exception will inform the user of this function that something has gone wrong, logging an error doesn't provide an additional value, instead logger.info can be used to log meaningful information for the debugging activity. Also, if there are log-scanners (generally, there should be) to raise alarms/tickets on error log count, then we get notified of 1 error 2 times - once through the thrown exception and second from log scanning.

code/backend/batch/BatchPushResults.py
```diff
@@ -27,21 +27,38 @@ def _get_file_name_from_message(msg: func.QueueMessage) -> str:
     arg_name="msg", queue_name="doc-processing", connection="AzureWebJobsStorage"
 )
 def batch_push_results(msg: func.QueueMessage) -> None:
-    do_batch_push_results(msg)
+    message_body = json.loads(msg.get_body().decode("utf-8"))
+    logger.info("Process Document Event queue function triggered: %s", message_body)
```
gaurarpit, line 31: for consistency can we please have f-strings in all places ? we have a scattered usage of regular strings ( like this line) and fstrings ( like line 45) ?

gaurarpit (commented):
nice, looks good. some comments.

code/backend/batch/BatchPushResults.py
liammoat: I generally use `printf`-style %-formatting with loggers. Mainly for support with structured loggers like Sentry, but also for perf. For reference, nice explanation [here](https://blog.pilosus.org/posts/2020/01/24/python-f-strings-in-logging/).

However, I am happy to standardise either way. @ross-p-smith @adamdougal - is there a project preference?

code/backend/batch/BatchPushResults.py
liammoat: good spot - thanks, will update

code/backend/batch/BatchPushResults.py
ross-p-smith: I don't believe there is a preference - I'm happy as long as we are consistent. When we decide, maybe a line in the CONTRIBUTING.md - we use " . . . "

code/backend/batch/function_app.py
```diff
@@ -10,7 +10,8 @@
 logging.captureWarnings(True)
 # Raising the azure log level to WARN as it is too verbose - https://github.com/Azure/azure-sdk-for-python/issues/9422
 logging.getLogger("azure").setLevel(os.environ.get("LOGLEVEL_AZURE", "WARN").upper())
-configure_azure_monitor()
+if os.getenv("APPLICATIONINSIGHTS_ENABLED", "false").lower() == "true":
```
superhindupur, line 13: Can we use EnvHelper's APPLICATIONINSIGHTS_ENABLED field instead? Could save us from the .lower() and == "true" logic. @adamdougal do you see any issues with instantiating env helper in this file?

superhindupur:
> Team - Couple of things to note about this PR:
> 
> 1. There is likely a better way to do this - however, I have tried to implement it with as few downstream changes as possible (i.e. I don't want to refactor everything to fix this)
> 2. With this change in mind, I propose the admin app should delete the blob, and let the function handle the index.
> 3. I will leave this in draft until I have introduce some tests to cover the core behaviour

Definitely agree with " I propose the admin app should delete the blob, and let the function handle the index." - could you please create an issue to cover this if there isn't one already?

superhindupur:
@liammoat thanks so much for this, really appreciate it! ❤️

code/backend/batch/function_app.py
adamdougal: Ah so we did that before. But then the `EnvHelper` gets initialised, with all of it's calls to KeyVault etc **before** application insights is configured, meaning we lose visibiliy on those calls :(

code/backend/batch/BatchPushResults.py
adamdougal: I'm happy either way. I'm not 100% familiar with the python logger, but loggers do sometimes do some nice formatting when using this approach

adamdougal (dismissed):
LGTM

gaurarpit (approved):

code/backend/batch/function_app.py
liammoat: @adamdougal - do we need visibility of those calls? Particularly if the projects switches to Key Vault references (see #625)

adamdougal: I think it'd be good to keep it, and then revaluate after #625. We weren't really aware of the number of KeyVault calls we were making until we added this, and turned out it was quite a lot!

cecheta: I think another issue is that you are meant to call `configure_azure_monitor()` as early as possible, before importing any of the modules that you want to instrument, because the `configure_azure_monitor()` function essentially "monkeypatches" the modules to add instrumentation

cecheta: But it doesn't look like we're doing that currently anyway 🤷‍♂️

code/backend/batch/utilities/search/SearchHandlerBase.py
```diff
@@ -36,13 +36,36 @@ def get_files(self):
...
+    def delete_by_source(self, source) -> None:
+        if source is None:
+            return
+
+        documents = self._get_documents_by_source(source)
+        if documents is None:
+            return
+
+        results = self.output_results(documents)
+        files_to_dete = {filename: ids for filename, ids in results.items()}
```
cecheta, line 59: What does this line do?

code/tests/test_BatchPushResults.py
```diff
@@ -15,17 +16,19 @@
 
 @pytest.fixture(autouse=True)
 def get_processor_handler_mock():
-    with patch("backend.batch.BatchPushResults.EmbedderFactory.create") as mock:
-        processor_handler = mock.return_value
-        yield processor_handler
+    with patch("backend.batch.BatchPushResults.EmbedderFactory.create") as mock_create_embedder, \
```
cecheta, line 19: Could you get pre-commit up and running so it formats the modified files correctly?

code/backend/batch/function_app.py
liammoat: Thanks for the discussion team. I have opted to keep it, and then revaluate after #625

code/tests/test_BatchPushResults.py
liammoat: @cecheta - this has ran through pre-commit.

The only files to fail are outside the diff of this PR:
image

What format is this missing?

cecheta: We don't usually use `\` for new lines, but if `black` is happy with it then that's fine, sorry

code/backend/batch/BatchPushResults.py
```diff
@@ -7,14 +7,14 @@
...
+from utilities.search.Search import Search
 
 bp_batch_push_results = func.Blueprint()
 logger = logging.getLogger(__name__)
 logger.setLevel(level=os.environ.get("LOGLEVEL", "INFO").upper())
 
 
-def _get_file_name_from_message(msg: func.QueueMessage) -> str:
-    message_body = json.loads(msg.get_body().decode("utf-8"))
+def _get_file_name_from_message(message_body: any) -> str:
```
cecheta, line 17: My editor is complaining that these types should be `Any`?

ross-p-smith: str | byte  ?

code/backend/batch/utilities/search/SearchHandlerBase.py
liammoat: Good spot @cecheta. On it

code/backend/batch/BatchPushResults.py
liammoat: `Any` would require an import from `typings`: 

image

VSCode, Pycharm, and Lance are all happy with `any` for @gaurarpit and I

cecheta: It thinks it's the `any()` function

liammoat: updated now following discussion with @cecheta

cecheta (dismissed):
Great work

.vscode/settings.json
```diff
@@ -4,6 +4,7 @@
     "azureFunctions.scmDoBuildDuringDeployment": true,
     "azureFunctions.projectLanguage": "Python",
     "azureFunctions.projectRuntime": "~4",
+    "azureFunctions.pythonVenv": "${workspaceFolder}/.venv",
```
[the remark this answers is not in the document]
liammoat: When debugging Azure Functions, `func` would select the wrong `python` - and not have the modules etc.

For ref: https://github.com/Azure/azure-functions-core-tools/issues/3042

This change resolved the issue.  However, it is temperamental either way. So, I'll remove it for now

cecheta (approved):

PR #904, opened 2024-05-14
ci: use GitHub token

frtibble:
## Purpose
This updates the Action which creates a release to:
- Use the GitHub token rather that a PAT token, this means that releases will be created as the github-actions bot rather than as the user who generated the PAT
- This requirement made it difficult to use `release-please` so I have opted to instead use the [conventional-changelog-action](https://github.com/TriPSs/conventional-changelog-action). This has the same features (generates changelog, version number), however it allows us to publish the release directly, rather than creating a PR. This is important because the GitHub token does not have permissions to raise a PR on our behalf. 
- Note that, this does not mean releases will be published on every merge to main. They will only be published when a change results in a version increase. 
- It also means that release PRs do not need to be approved (this may be viewed as a good or a bad thing - we can discuss). 

- [ ] Yes
- [x] No


gaurarpit (dismissed):

ross-p-smith (approved):

frtibble:
@gaurarpit , @ross-p-smith, @adamdougal - have pushed a few changes, this takes a slightly different approach so please see the description of the PR for an update. Happy to discuss in more detail tomorrow - this feels to me a cleaner solution than before.

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,37 @@
+on:
+  push:
+    branches:
+      - main
```
cecheta, line 4: Do you think it would be possible to run this after the CI job succeeds?

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,37 @@
...
+
+jobs:
+  create-release:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Conventional Changelog Action
```
cecheta, line 20: Is this going to commit directly on `main`? Will it be able to?

package.json
```diff
@@ -0,0 +1,8 @@
+{
```
cecheta, line 1: Could we update the `pyproject.toml` instead of adding a `package.json`?

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,37 @@
...
+
+name: create-release
+
+jobs:
+  create-release:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
```
cecheta, line 18: We can use a newer checkout version

frtibble: Nice spot 💪

package.json
frtibble: Done! Have added `version` into the `pyproject.toml` and deleted package.json

.github/workflows/create-release.yml
frtibble: Yeah I think this is a great suggestion, I've updated it so that it is now triggered on CI workflow, and will only create a release if the CI succeeds. That was based on my assumption that the CI action only gets run on merges to main, not for dev, etc.

.github/workflows/create-release.yml
frtibble: It should commit to main yes, I've been testing it in a personal repo which has the same permissions set and it is able to. 

As to why/how, since it has 

```
permissions:
  contents: write
```

that allows the action to create a release (from [here](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#overview). So I _think_ its ability to push a commit to `main` is with the restriction that it's a release PR and not introducing any additional code. There's a separate (currently disabled) permission that would allow an Action to be able to merge PRs using the GitHub token. 

Speaking of, I think I can remove the permission: pull requests since that's not needed here

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Conventional Changelog Action
+        id: changelog
+        uses: TriPSs/conventional-changelog-action@v5
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          output-file: 'CHANGELOG.md'
+          git-user-name: 'github-actions[bot]'
+          git-user-email: 'github-actions[bot]@users.noreply.github.com'
+          version-file: './pyproject.toml'
```
cecheta, line 29: If you set `version-path` to `tool.poetry.version`, would it mean that we don't have to add a new `version` top-level field?

cecheta: I of course haven't tried this myself, just looking at the docs

frtibble: yes it does, just tried this out and it works a charm!

cecheta (approved):
Looking forward to seeing this in action 😀

PR #912, opened 2024-05-15
fix: Keep the Admin.py as uppercase naming to allow the streamlit pick it as is.

gaurarpit:
## Purpose

* Fix the application right now by having Admin.py as is.
* 

- [ ] Yes
- [x] No


cecheta (changes requested):
Could you change the PR title to be more descriptive, it will end up on the changelog

gaurarpit:
> Could you change the PR title to be more descriptive, it will end up on the changelog

Updated. thanks. good shout.

cecheta (approved):
Thanks

superhindupur (approved):

PR #919, opened 2024-05-15
Fix computer vision for deployments

cecheta:
## Purpose

* This PR handles the conditional creation of the computer vision resource
* Removes `--insert-final-newline` from the `generate_arm_templates.sh` script, as:
  * It doesn't seem to be needed anymore
  * It was failing silently

- [ ] Yes
- [x] No

## How to Test
Deploy different configurations, with/without computer vision


cecheta:
I am going to separate the re-formatting to a new PR

komalg1 (approved):

superhindupur (approved):
So the problem was that we were assuming the presence of computer vision parameters?

cecheta:
> So the problem was that we were assuming the presence of computer vision parameters?

Correct, but the module was being created conditionally

Error messages weren't clear at all

adamdougal (approved):
Makes sense! I should have tested with advanced image processing disabled :(

PR #921, opened 2024-05-15
feat: Store image embeddings in search index

adamdougal:
- hard coded the vector search dimensions for now
  - once we have implemented the `vectorizeText` call when searching, we
    could do the same thing we've done for GPT embeddings


code/backend/batch/utilities/helpers/azure_search_helper.py
```diff
@@ -121,6 +121,18 @@ def create_index(self):
...
         ]
 
+        if self.env_helper.USE_ADVANCED_IMAGE_PROCESSING:
+            logger.info("Adding image_vector field to index")
+            fields.append(
+                SearchField(
+                    name="image_vector",
+                    type=SearchFieldDataType.Collection(SearchFieldDataType.Single),
+                    searchable=True,
+                    vector_search_dimensions=1024,
```
gaurarpit, line 131: thinking aloud: is this something which we want as a constant or a config ?
If constant, can we make it a class level or method level constant ?  Or if config does it also move to ENV ?

adamdougal: So I think in future we want to "look up" this value similar to what we do for the `content_vector` dimensions https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/main/code/backend/batch/utilities/helpers/azure_search_helper.py#L71-L76.

For this though we need to implement the call to the Computer Vision `vectorizeText` end point. That's not something we _need_ until we implement querying.

I suggest once we have that implemented we revist this and dynamically set this field.

gaurarpit: that's fair. thanks for letting me know.

ross-p-smith (approved):

gaurarpit (approved):

code/tests/utilities/helpers/test_push_embedder.py
```diff
@@ -112,35 +113,90 @@ def azure_computer_vision_mock():
...
+def test_embed_file_advanced_image_processing_raises_exception_on_failure(
+    azure_search_helper_mock,
+):
+    # given
+    push_embedder = PushEmbedder(MagicMock(), MagicMock())
+
+    successful_indexing_result = MagicMock()
+    successful_indexing_result.succeeded = True
+    failed_indexing_result = MagicMock()
+    failed_indexing_result.succeeded = False
```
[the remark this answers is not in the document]
adamdougal: Did not know you could do that! Lets update in the next PR!

PR #922, opened 2024-05-15
fix: Fix generate_arm_templates.sh script, reformat all Bicep files

cecheta:
## Purpose

* Fixes the `generate_arm_templates.sh` script, which was failing silently
* Format all Bicep files

There are no actual changes to the infra.

- [ ] Yes
- [x] No


gaurarpit (approved):
looks good, how do we test this ?

cecheta:
> looks good, how do we test this ?

I quite crudely ran this:

```bash
for FILE in $(find . -name '*.bicep'); do
  echo '' >> $FILE
done
```

Then I made a new commit

Issue #864, opened 2024-05-09
Automatically check the PEP8 compliance. TBD - Done after the changes.


gaurarpit:
Required by #514 

* Automatically check the PEP8 compliance. TBD - Done after the changes.

gaurarpit:
we have decided against this. All the files except "Admin.py" are snake case and we'll not add automation to this anymore. 
We will rely on code review process to ensure that the naming stays snake_case. Closing this task.

PR #925, opened 2024-05-15
style: Investigate pylint for PEP8

ross-p-smith:
## Purpose
Wanted to investigate the benefit of using pylint which seems to give us better vscode integration.

This PR is more for discussion as it relates #864

### DIFF
--- a/code/app.py
+++ b/code/app.py
@@ -1,3 +1,7 @@
+"""
+This module contains the entry point for the application.
+"""
+
 import os
 import logging
 from azure.monitor.opentelemetry import configure_azure_monitor
@@ -5,7 +9,8 @@ from opentelemetry.instrumentation.httpx import HTTPXClientInstrumentor
 
 logging.captureWarnings(True)
 logging.basicConfig(level=os.environ.get("LOGLEVEL", "INFO").upper())
-# Raising the azure log level to WARN as it is too verbose - https://github.com/Azure/azure-sdk-for-python/issues/9422
+# Raising the azure log level to WARN as it is too verbose -
+# https://github.com/Azure/azure-sdk-for-python/issues/9422
 logging.getLogger("azure").setLevel(os.environ.get("LOGLEVEL_AZURE", "WARN").upper())
 # We cannot use EnvHelper here as Application Insights should be configured first
 # for instrumentation to work correctly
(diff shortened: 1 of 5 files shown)


code/create_app.py
```diff
@@ -180,45 +187,51 @@ def conversation_with_data(request: Request, env_helper: EnvHelper):
         }
 
         return response_obj
-    else:
-        return Response(
-            stream_with_data(response),
-            mimetype="application/json-lines",
-        )
+
```
ross-p-smith, line 190: https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/no-else-return.html

code/create_app.py
```diff
@@ -271,10 +285,8 @@ def conversation_without_data(request: Request, env_helper: EnvHelper):
             ],
         }
         return jsonify(response_obj), 200
-    else:
-        return Response(
-            stream_without_data(response), mimetype="application/json-lines"
-        )
+
```
ross-p-smith, line 288: https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/no-else-return.html

code/app.py
```diff
@@ -1,18 +1,24 @@
...
+# https://github.com/Azure/azure-sdk-for-python/issues/9422
 logging.getLogger("azure").setLevel(os.environ.get("LOGLEVEL_AZURE", "WARN").upper())
 # We cannot use EnvHelper here as Application Insights should be configured first
 # for instrumentation to work correctly
 if os.getenv("APPLICATIONINSIGHTS_ENABLED", "false").lower() == "true":
     configure_azure_monitor()
     HTTPXClientInstrumentor().instrument()  # httpx is used by openai
 
+# pylint: disable=wrong-import-position
 from create_app import create_app  # noqa: E402
```
gaurarpit, line 22: just realized, shouldn't this move to the top of the file ? ( we did that story a fews day back, right ? )

code/create_app.py
gaurarpit: ahaa.. this is so nice. My pet peeve is when else block literally has only 1 line that returns.

gaurarpit (approved):

code/app.py
ross-p-smith: This is for this 1 particular line - this is so we setup logging before we create the app, but also allows us to test the creat_app code without all the logger infrastructure

Issue #866, opened 2024-05-09
DevContainer doesn't work on Mac.


gaurarpit:
# Describe the bug
DevContainer doesn't work on Mac.
The image of python version needed inside the devcontainer doesn't build on ARM. Changing it causes the `azure-core-function` to break. 

# Expected behavior
The devcontainer should work on Mac devices with ARM processor.

# How does this bug make you feel?
_Share a gif from [giphy](https://giphy.com/) to tells us how you'd feel_

---

# Debugging information

## Steps to reproduce
1. Get a Mac ARM processor laptop
2. Open the repository in devcontainer on VSCode Code Editor
3. Do `azd provision` and the below error is thrown

## Screenshots

![SCR-20240508-lodo](https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/assets/4739534/a1aef773-fe29-4a9e-9032-ae164e0ba13c)

gaurarpit:
Root cause seems to be this: https://github.com/Azure-Samples/azure-search-openai-demo/issues/195#issuecomment-1552219371

We suspect that the `azurebicep` OR `azure-functions-core-tools` installer might be the reason.

ross-p-smith:
@gaurarpit or @liammoat  - are you able to pull my branch and see if it builds on your Mac?

liammoat:
> @gaurarpit or @liammoat - are you able to pull my branch and see if it builds on your Mac?

Previously, the dev container would launch - but fail with `azd`. With your branch, the dev container image now fails here:
![image](https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/assets/2767292/0085c303-ee36-4e47-809c-df6121ba79cf)

mikqs:
Would you have a workaround for this bug?

mikqs:
I have the latest merge and an M3 Mac
```
=> ERROR [dev_containers_target_stage 8/9] RUN --mount=type=bind,from=de  2.7s
------
[...]

PR #926, opened 2024-05-16
build: Bullseye -> bookworm

ross-p-smith:
## Purpose
There was a problem with the Functions Core Tools that the compression was not working.  This meant that we had to tag the version we used.

By not defining that we have to run on Bullseye and use Bookworm (debian 12) instead, we no longer need to pin that version. Testing that this works on Windows first and then seeing if we can get Mac to work on top of this.

Relates to #866 

Please bear in mind that you will need to "rebuild" your devcontainer

After:
```sh
❯ cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
NAME="Debian GNU/Linux"
VERSION_ID="12"
VERSION="12 (bookworm)"
VERSION_CODENAME=bookworm
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
❯ azd version
azd version 1.9.2 (commit c58b02f71710960aba28f81f3698e64cfdda9f96)
❯ func version
4.0.5700
```

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,5 +1,4 @@
-FROM --platform=linux/amd64 mcr.microsoft.com/devcontainers/python:1-3.11-bullseye
-# We need to force the container to be amd so that it works on a Mac. Without this the functions extension doesn't install.
+FROM mcr.microsoft.com/devcontainers/python:3.11
 
 # install git
 RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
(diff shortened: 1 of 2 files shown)


adamdougal (approved):
lgtm

gaurarpit (approved):
(fingers crossed - this might fix the mac issue)

PR #933, opened 2024-05-16
fix: Cache the Config 

### LINKED ISSUE #647: Cache config in application
# Motivation

Every time we call `ConfigHelper.get_active_config_or_default()` we make a request to the storage account, which happens multiple times per request. Considering that this is not updated often, we should make this request once and cache the value in memory. We can add a note that the app will need to be restarted if the config is updated, in the same way that updating environment variables causes the application to restart.

# Requirements

- Fetch config once during application startup
- Save the config within the class/module
- Update message to mention restarting app after updating config https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/3e71fe10c50a129a20c314ba015065acdc6d8379/code/backend/pages/04_Configuration.py#L212

# Tasks

ross-p-smith:
## Purpose
Fixes #647 

We use the higher-order functions to decorate the `get_active_config_or_default` method with a cache. We also invalidate the cache when we save the config.

However, because there are 2 applications, we display some text to tell the user that they will need to restart the Chat App to invalidate the cache


.vscode/settings.json
```diff
@@ -21,5 +21,4 @@
     "python.testing.cwd": "${workspaceFolder}/code",
     "python.testing.unittestEnabled": false,
     "python.testing.pytestEnabled": true,
-    "pylint.path" : [ "${interpreter}", "-m", "pylint" ]
```
ross-p-smith, line 24: This prevents pylint from working in the DevContainer - we should find a solution for both. @liammoat - Assume you added this to get this working in a Mac

adamdougal (dismissed):
Looks great! Just wondering, how much effort do you think it would be to add an endpoint to the chat app to reset/clear the config that we call after saving? The caveat being that if their are multiple instances of the chat app it won't work. But not sure if that's something we allow to be configured. WDYT?

lgtm

ross-p-smith:
> how much effort do you think it would be to add an endpoint to the chat app

I was actually thinking about this. However, I think the problem would be the dependency graph of the deployment. We would need the chat app deployed to be able to get the url (unless we did it by convention), but the backend is what is deployed last; and it would be the admin app calling the backend. But not saying no, let me give it a weekend thought!

adamdougal:
> > how much effort do you think it would be to add an endpoint to the chat app
> 
> I was actually thinking about this. However, I think the problem would be the dependency graph of the deployment. We would need the chat app deployed to be able to get the url (unless we did it by convention), but the backend is what is deployed last; and it would be the admin app calling the backend. But not saying no, let me give it a weekend thought!

Ah good point! I guess another option could be to have a ttl on the cache, is that possible with func tools?

code/backend/pages/04_Configuration.py
```diff
@@ -379,9 +379,11 @@ def validate_documents():
             ),
         }
         ConfigHelper.save_config_as_active(current_config)
-        st.success("Configuration saved successfully!")
+        st.success(
+            "Configuration saved successfully! Please restart the chat service for these changes to take effect."
+        )
 
-    with st.popover(":red[Reset confiiguration to defaults]"):
+    with st.popover(":red[Reset configuration to defaults]"):
```
cecheta, line 386: 🙈

cecheta (dismissed):
Nice

adamdougal (approved):
LGTM!

PR #934, opened 2024-05-17
fix: Add `max_tokens` to chat completions

cecheta:
## Purpose

* GPT-4 has a default max tokens of 16, meaning the captions generated were very small. This PR adds the `AZURE_OPENAI_MAX_TOKENS` to the chat completions call, so it will default to 1000 tokens, and can be conifgured.

- [ ] Yes
- [x] No

## How to Test
* Enable Advanced Image Processing
* Upload image
* View caption in Azure Search

### DIFF
--- a/code/backend/batch/utilities/helpers/llm_helper.py
+++ b/code/backend/batch/utilities/helpers/llm_helper.py
@@ -30,7 +30,7 @@ class LLMHelper:
 
         self.llm_model = self.env_helper.AZURE_OPENAI_MODEL
         self.llm_max_tokens = (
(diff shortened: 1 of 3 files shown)


adamdougal:
Do we know why that's the default? Is it more expensive?

cecheta:
> Do we know why that's the default? Is it more expensive?

Not sure, seems to be just for vision

* https://community.openai.com/t/documented-max-token-default-is-incorrect-for-gpt-4-vision-preview/507329
* https://community.openai.com/t/gpt-4-vision-preview-finish-details/475911/6

adamdougal (approved):
LGTM!

PR #936, opened 2024-05-17
fix: Error viewing images in Explore Data page

cecheta:
## Purpose

* Fixes error where images uploaded with advanced image processing do not display in the Admin app, as they do not have a chunk ID in their metadata
* The fix is to default to chunk ID 0 when rednering the data

- [ ] Yes
- [x] No

## How to Test
* Enable advanced image processing
* Upload image
* View data in Explore Data page

image


adamdougal (approved):
lgtm

Issue #750, opened 2024-04-24
Search image data from `/api/conversation/custom`


adamdougal:
Required by https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/issues/323

# Description

When calling the `/api/conversation/custom` endpoint, generate embeddings of the question using computer vision and query the image embeddings using the question embeddings. We should continue to create question embeddings using `text-embeddings-ada-002` in addition to be used to search the image description.

# Tasks

- [x] Stop using langchain in the quesiton answer tool and post prompt tool
- [x] Generate question embeddings using computer vision
- [x] Query the search index using the generated embeddings
- [x] Pass any returned images to `gpt-4-vision` along with the chat
- [x] Add a limit on the number of images passed to vision

PR #939, opened 2024-05-17
feat: Remove LangChain from question answer tool

cecheta:
Required for #750

## Purpose

* This PR removes LangChain from the Question Anwer Tool
* Changes the example user/assistant messages to system messages, as from the [docs](https://cookbook.openai.com/examples/how_to_format_inputs_to_chatgpt_models#few-shot-prompting)
* Updates unit and functional tests

- [ ] Yes
- [x] No

## How to Test
* Ask questions to LLM that require Question Answer Tool


adamdougal (dismissed):
Just one question about the temperature but looks great!

great

code/tests/functional/tests/backend_api/default/test_conversation_custom.py
```diff
@@ -559,11 +561,8 @@ def test_post_makes_correct_call_to_openai_chat_completions_with_documents(
                         "role": "user",
                     },
                 ],
-                "model": "gpt-3.5-turbo",  # This is hardcoded in LangChain
+                "model": app_config.get("AZURE_OPENAI_MODEL"),
                 "max_tokens": int(app_config.get("AZURE_OPENAI_MAX_TOKENS")),
-                "n": 1,
-                "stream": False,
-                "temperature": float(app_config.get("AZURE_OPENAI_TEMPERATURE")),
```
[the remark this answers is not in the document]
cecheta: Good point, I've added it back

I'm not passing in temperature as `None`, as apparently using `None` can have different results to leaving it blank. I also didn't want to hardcode it to `0` as the `get_chat_completion` function is being used in multiple places

code/backend/batch/utilities/tools/question_answer_tool.py
```diff
@@ -124,47 +110,52 @@ def generate_on_your_data_llm_chain(
...
 
         llm_helper = LLMHelper()
 
-        answer_generator = LLMChain(
-            llm=llm_helper.get_llm(), prompt=answering_prompt, verbose=self.verbose
-        )
-
-        with get_openai_callback() as cb:
-            result = answer_generator(input)
+        response = llm_helper.get_chat_completion(messages, temperature=0)
```
adamdougal, line 148: Should this be:

```suggestion
        response = llm_helper.get_chat_completion(messages, temperature=self.env_helper.AZURE_OPENAI_TEMPERATURE)
```
?

cecheta: It probably should be, however I've tried to keep it the same as before. In LangChain it's hardcoded to `0`.

`AZURE_OPENAI_TEMPERATURE` is only used in the BYOD endpoint

adamdougal: Gotcha, I'm thinking we should probably set that now that we have control. The env var defaults to 0 so shouldn't cause any issues.

cecheta: We had control before, we just set it to 0

https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/54723fe9b13baf1096abff135c67a98fcdf1a2cb/code/backend/batch/utilities/helpers/llm_helper.py#L43

I'll keep it as 0 for now to reduce the changes from what should essentially be a refactor, but will raise a new issue for this

cecheta: #941

adamdougal (approved):

PR #954, opened 2024-05-20
fix: EnvHelper unusable if error occurs during initialisation

cecheta:
## Purpose

* If an error occurs during env helper initialisation, the helper remains in a non-working state until the app restarts. This PR fixes this, so the env helper is only saved if it is created without any errors

- [ ] Yes
- [x] No

### DIFF
--- a/code/backend/batch/utilities/helpers/env_helper.py
+++ b/code/backend/batch/utilities/helpers/env_helper.py
@@ -15,8 +15,9 @@ class EnvHelper:
     def __new__(cls):
         with cls._lock:
             if cls._instance is None:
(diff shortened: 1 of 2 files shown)


ross-p-smith:
Not for this PR - but wonder whether the `@functools.cache` would be simpler than the lock

ross-p-smith (approved):

cecheta:
> Not for this PR - but wonder whether the `@functools.cache` would be simpler than the lock

From what I quickly saw online, there could be multiple threads trying to create EnvHelpers at the same time during startup unless we use a lock

I don't think that would break the application, but I guess it wouldn't be a true singleton

PR #961, opened 2024-05-22
style: Create a pre-commit hook for checking snake casing of files.

gaurarpit:
## Purpose

* Create a pre-commit hook for checking snake casing of files

- [ ] Yes
- [x] No

## How to Test
* Tested locally by manually changing file names and casing.


ross-p-smith:
I thought we spoke about this in Sprint Planning that this has no value.

gaurarpit:
> I thought we spoke about this in Sprint Planning that this has no value.

it was already done.

adamdougal:
I think it's probably worth leaving this out for now, maybe we can add it in if we start active work on this project again.

gaurarpit:
> I think it's probably worth leaving this out for now, maybe we can add it in if we start active work on this project again.

Sure, I agree with you. I'll abandon this PR for now.

PR #966, opened 2024-05-22
fix: Add `vector_fields` to Azure OpenAI On Your Data

cecheta:
## Purpose

* Fixes BYOD endpoint, which was not working when using data

- [ ] Yes
- [x] No

## How to Test
* `azd env set CONVERSATION_FLOW byod`
* Ask question that involves data

## What to Check
* Response received


ross-p-smith (approved):
in1

Issue #964, opened 2024-05-22
Include image citations in prompt/response


cecheta:
Required by #323 

# Description

When passing images to GPT-4 vision in the question answer tool, the prompt should be modified if advanced image processing is enabled, so that the LLM is instructed to include citations to the image in the response, and the response is parsed as appropriate.

# Tasks

_To be filled in by the engineer picking up the subtask

PR #971, opened 2024-05-23
feat: Include images in call to LLM

cecheta:
Required for #750

## Purpose

* Pass retrieved images to LLM when advanced image processing is enabled

- [ ] Yes
- [x] No

## How to Test
* Run tests
* Deploy with advanced image processing enabled, upload images, ask questions based on images


code/backend/batch/utilities/tools/question_answer_tool.py
```diff
@@ -122,20 +125,41 @@ def generate_on_your_data_messages(
             },
             *QuestionAnswerTool.clean_chat_history(chat_history),
             {
-                "content": self.config.prompts.answering_user_prompt.format(
-                    sources=documents,
-                    question=question,
-                ),
+                "content": [
```
cecheta, line 128: This format works even when not using vision

cecheta:
Note that we ran into this issue where the LLM attempts to provide an answer from every image passed to the LLM, even if it's not relevant

Screenshot 2024-05-21 145846

This should be looked into during #964

adamdougal:
> Note that we ran into this issue where the LLM attempts to provide an answer from every image passed to the LLM, even if it's not relevant
> 
> Screenshot 2024-05-21 145846
> 
> This should be looked into during #964

Is this something to do with `top_k`? Or is this something we can tweak in the system prompt to say ignore anything not relevant?

adamdougal (dismissed):
lgtm

code/backend/batch/utilities/tools/question_answer_tool.py
```diff
@@ -145,8 +169,39 @@ def answer_question(self, question: str, chat_history: list[dict], **kwargs):
...
+        )
+
+        return clean_answer
+
+    def create_image_url_list(self, source_documents):
+        image_types = self.config.get_advanced_image_processing_image_types()
+
+        image_documents = []
+
+        for doc in source_documents:
```
ross-p-smith, line 184: Apologies, I'm reading this in the browser. But it seems we iterate around one list and then iterate around it again to add the sas token - could we do this all in one iteration?

cecheta: Good point, done

cecheta:
> > Note that we ran into this issue where the LLM attempts to provide an answer from every image passed to the LLM, even if it's not relevant
> > Screenshot 2024-05-21 145846
> > This should be looked into during #964
> 
> Is this something to do with `top_k`? Or is this something we can tweak in the system prompt to say ignore anything not relevant?

I think it would need to be done in the prompt

adamdougal (approved):

PR #976, opened 2024-05-24
fix: PSL-BUG-1850-Fixed

Prasanjeet-Microsoft:
## Purpose
Web app - Part listed as reference is not included in Citation preview panel and after fixing this bug part no will give proper citation preview panel on the right side of screen.

- [ ] Yes
- [x] No

## How to Test
* If you type something in chatbot for eg. "Describe in more detail the risks from market volatility" we are getting response with the citation listed as Parts 19 and 14, respectively. This "Part" of the citation is not included in the Citation preview panel.

## What to Check
* After receiving response in chatbot click on reference link which contains part no, that part no should include in the Citation preview panel.


hunterjam (approved):
Change approved.

Roopan-Microsoft (approved):
Approved the changes.

PR #977, opened 2024-05-24
fix: delete url embeddings from index when deleting from Admin website

### LINKED ISSUE #975: Unable to delete URLs from knowledge base
# Describe the bug
Following #970, it is no longer possible to delete data from a URL through the Admin app, as there is no corresponding file in Azure Storage

# Expected behavior
It still be possible to delete data that has been ingested from a URL, via the Admin app

---

# Debugging information

## Steps to reproduce
* Deploy application
* Upload data via URL
* Attempt to delete data via Admin app
* The data is still present

komalg1:
## Purpose

* Resolves #975 
* Deleting from index every time user deletes from Admin site

- [ ] Yes
- [x] No

## What to Check
* Deleting url or documents from Admin website deletes it from storage as well as index
* Deleting document directly from Blob Storage deletes the documents from index

### DIFF
--- a/code/backend/batch/utilities/search/search_handler_base.py
+++ b/code/backend/batch/utilities/search/search_handler_base.py
@@ -24,6 +24,8 @@ class SearchHandlerBase(ABC):
 
     def delete_from_index(self, blob_url) -> None:
         documents = self.search_by_blob_url(blob_url)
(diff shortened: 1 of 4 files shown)


cecheta (approved):
Nice

PR #994, opened 2024-05-28
ci: Group langchain updates together

cecheta:
## Purpose

* Group langchain updates together, as one cannot be updated without the other

- [ ] Yes
- [x] No


adamdougal (approved):
Nice! Did not know you could do this!

PR #1002, opened 2024-05-28
fix: Frontend crash due to invalid citations

### LINKED ISSUE #707: Frontend crash due to invalid citations
# Describe the bug
The frontend crashed after receiving a response from the `/custom` endpoint. I was not able to obtain the response, however the crash occurred at this line, as `citation` was `undefined`:

https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/f755b8ec3c30c532347b456b49a1d5d48f80cf77/code/frontend/src/components/Answer/AnswerParser.tsx#L24

The error most likely occurred because the response contained a citation that did not exist, i.e. `[doc6]`, even though there were only 5 source documents.

The following was observed in the API logs:

```
Source document 6 not provided, skipping doc
Source document 7 not provided, skipping doc
Source document 8 not provided, skipping doc
Source document 9 not provided, skipping doc
```

# Expected behavior

The frontend application should not crash. Errors should be caught and an handled gracefully, or an appropriate error is returned to the user.

---

# Debugging information

## Steps to reproduce

Difficult to reproduce due to non-deterministic result.

## Logs

Provided above.

---

# Tasks

komalg1:
## Purpose

* Fixes #707 

- [ ] Yes
- [x] No

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

## What to Check
* Front end should not crash in case the citation does not exist for a source document.


code/frontend/src/components/Answer/AnswerParser.tsx
```diff
@@ -19,7 +19,7 @@ export function parseAnswer(answer: AskResponse): ParsedAnswer {
         // Replacing the links/citations with number
         let citationIndex = link.slice(lengthDocN, link.length - 1);
         let citation = cloneDeep(answer.citations[Number(citationIndex) - 1]) as Citation;
-        if (!filteredCitations.find((c) => c.id === citationIndex)) {
+        if (!filteredCitations.find((c) => c.id === citationIndex) && citation != undefined) {
```
cecheta, line 22:
```suggestion
        if (!filteredCitations.find((c) => c.id === citationIndex) && citation !== undefined) {
```

cecheta (approved):
Thanks for fixing

PR #1004, opened 2024-05-29
fix: Correct the image_url for calling gpt vision

cyberflying:
Correct the image_url for calling gpt vision

## Purpose

* When the use_advanced_image_processing option is enabled, the GPT vision model is invoked to generate the image caption, and the image_url parameter needs to be converted to the format of the data url.

- [ ] Yes
- [x] No

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

## Other Information

backend function error log:
image


adamdougal:
Heya @cyberflying! Thanks for this PR.

As per https://platform.openai.com/docs/guides/vision?lang=curl you can supply either a url or a base64 encoded verison of the image. Based on that, I don't think this is the cause of your issue. The logs you've provided just show that the function has failed, there should be other logs which show what exactly caused that failure. Can you provide those? 

Thanks

cyberflying:
> Heya @cyberflying! Thanks for this PR.
> 
> As per https://platform.openai.com/docs/guides/vision?lang=curl you can supply either a url or a base64 encoded verison of the image. Based on that, I don't think this is the cause of your issue. The logs you've provided just show that the function has failed, there should be other logs which show what exactly caused that failure. Can you provide those?
> 
> Thanks

Hi @adamdougal ,
Thanks for your reply!

I using gpt-4o and openai==1.30.4, if using a url directly, then encounter error:
raise self._make_status_error_from_response(err.response) from None
openai.BadRequestError: Error code: 400 - {'error': {'message': "Invalid type for 'messages[1].content[1].image_url': **expected an object, but got a string instead."**, 'type': 'invalid_request_error', 'param': 'messages[1].content[1].image_url', 'code': 'invalid_type'}}

adamdougal:
Gotcha! So looking at that error message, and after taking a closer look at the docs. We need to change the request body from:
```json
{"image_url": source_url, "type": "image_url"},
```
to:
```json
{"image_url": {"url": source_url}, "type": "image_url"},
```

Are you happy to make that change and update the relevant tests?

adamdougal:
@cyberflying The change looks good! But tests are failing, could you update those too to expect this new structure? Thanks

cyberflying:
thanks @adamdougal ,  I've tested this change in my local environment, but I'm not familiar with how to test it in GitHub. Can you explain how to conduct the test?
I have checked some pr before. Is the test code like this:

cecheta:
Would it be possible to make the same change here?

https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/main/code/backend/batch/utilities/tools/question_answer_tool.py#L138-L142

adamdougal:
@cyberflying We have unit tests that you can run using the commands `make unittest functionaltest` that will run all tests.

Specifically it's this test that is failing https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/4b13f13069be4e5badcd7e2c8e702f3a8e216a22/code/tests/functional/tests/functions/advanced_image_processing/test_advanced_image_processing.py#L195-L197

cyberflying:
> Would it be possible to make the same change here?
> 
> https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/main/code/backend/batch/utilities/tools/question_answer_tool.py#L138-L142

ok, and done.

cyberflying:
> @cyberflying We have unit tests that you can run using the commands `make unittest functionaltest` that will run all tests.
> 
> Specifically it's this test that is failing
> 
> https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/4b13f13069be4e5badcd7e2c8e702f3a8e216a22/code/tests/functional/tests/functions/advanced_image_processing/test_advanced_image_processing.py#L195-L197

thanks! I'll try this unit test.

cyberflying:
@adamdougal ,  I have updated the tests code and passed the unit test and functional test.

cyberflying:
unit test: poetry run pytest -m "not azure and not functional"
results:
========================================================================================= short test summary info =========================================================================================
FAILED code/tests/utilities/helpers/test_env_helper.py::test_azure_speech_recognizer_languages_default - AssertionError: assert ['zh-CN', 'en-US'] == ['en-US']
FAILED code/tests/utilities/helpers/test_env_helper.py::test_use_advanced_image_processing[None-False] - assert True == False
FAILED code/tests/utilities/helpers/test_secret_helper.py::test_get_secret_returns_value_from_environment_variables - azure.core.exceptions.HttpResponseError: (BadParameter) The request URI contains an invalid name: my_secret_value
FAILED code/tests/utilities/helpers/test_secret_helper.py::test_get_secret_returns_empty_string_when_secret_name_is_empty - assert "[{'id': 'htt...eDays': 90}}]" == ''
================================================================== 4 failed, 212 passed, 73 deselected, 3 warnings in 166.39s (0:02:46) ===================================================================

functional test:  poetry run pytest code/tests/functional -m "functional"
results:
========================================================================================= short test summary info =========================================================================================
FAILED code/tests/functional/tests/backend_api/without_data/test_azure_byod_without_data.py::test_azure_byod_responds_successfully_when_streaming - assert 500 == 200
FAILED code/tests/functional/tests/backend_api/without_data/test_azure_byod_without_data.py::test_post_makes_correct_call_to_azure_openai - AssertionError: Matching request found 0 times but expected 1 times.
========================================================================== 2 failed, 60 passed, 2 warnings in 1429.29s (0:23:49) ==========================================================================

adamdougal:
Looks like some things are failing locally for you unexpectedly, we'll take a look into that. But your changes look good. Thanks for your contribution!

adamdougal (approved):

PR #1006, opened 2024-05-29
fix: backend .env path

djseng:
## Purpose

* use the same env_file the other two services use

- [ ] Yes
- [x] No

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
make docker-compose-up
```

## What to Check
* verify the backend can connect to the webjob storage as documented `docs/LOCAL_DEPLOYMENT.md#L114`

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -28,7 +28,7 @@ services:
     environment:
       - AzureWebJobsSecretStorageType=files
     env_file:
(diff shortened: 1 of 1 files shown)


adamdougal (approved):
LGTM! Thanks for the contribution!

PR #1008, opened 2024-05-31
fix: The Web UX is intentionally not revealing references when clicked to display

Rohini-Microsoft:
## Purpose
 The Web UX is intentionally not revealing references when clicked to display
 
 solution:  Added logic in Answer.tsx file in useEffect method. Also used useRef hook.

- [ ] Yes
- [x] No

## How to Test
* Tested locally by manually asked question in chatbot and will get references. When you got all references then click on the references link.

## What to Check
* .On click of references link then you will see all references.


adamdougal (approved):
Nice, LGTM!

Issue #973, opened 2024-05-23
Support image search in Integrated Vectorization


komalg1:
# Motivation

Currently Integrated Vectorization flow only supports text based search. By introducing a new OCR skill, image search can also be supported.
Stretch Goal - Using "Advanced Image Processing" can also be supported in Integrated Vectorization by using [Azure AI Vision Multimodal Embeddings Skill](https://learn.microsoft.com/en-us/azure/search/cognitive-search-skill-vision-vectorize)

# How would you feel if this feature request was implemented?

![gif](https://media.giphy.com/media/v1.Y2lkPTc5MGI3NjExN3ZoeW1jenFzYm94ZTZ3bWF1c3pkejdycGtmczhhOTk1Y2RkempxdSZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/0mKPytMujTds4e26IS/giphy.gif)

# Tasks

_To be filled in by the engineer picking up the issue_

- [ ] Task 1
- [ ] Task 2
- [ ] ...

PR #1021, opened 2024-06-04
feat: Integrated Vectorization - adding OCR skill

komalg1:
## Purpose

* This PR adds the OCR skill when using Integrated Vectorization to enable search on images.
* Relates to #973 

- [ ] Yes
- [x] No

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

## What to Check
* When Integrated Vectorization is enabled, user should be able to upload images with text and search those images


adamdougal:
I assume this still works fine for non image based data?

Do any functional tests need adding/updating for this?

adamdougal (approved):
Great work!

Issue #406, opened 2024-03-11
Include a prompt flow integration


adamdougal:
# Motivation

To demonstrate a lower code solution to chatting with your data. This could also be used **in future** to link with other experiment projects to auto generate the configuration for the flow.

# How would you feel if this feature request was implemented?

![flow](https://media.giphy.com/media/v1.Y2lkPTc5MGI3NjExOTNlaHM1MGczcGd3bnVzZmF0cWgya2RhMmxnN2Z4OG0xNG40bjluMCZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/h8y265b9iKtzKT0pDj/giphy.gif)

# Requirements

- Implement a basic version of chat with your data using prompt flow as the back end
- Allow the deployment/configuration of this via the current deployment mechanism
- Document this option exists and explain when to use it
- **Stretch:** implement all features of chat with your data using prompt flow

# Tasks

_To be filled in by the engineer picking up the issue_

- [x] #701
- [x] #915
- [x] #979 
- [x] Add vscode tooling for PromptFlow - https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/1040
- [x] Support citations (return chunk which was used to inform the answer) - https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/1089
- [x] Documentation

reechamishra:
@reechamishra  verify this with Randy if someone already working on this ? cc @prasbhargava1

reechamishra:
We need a design discussion to elaborate the requirements and acceptance criteria : @ross-p-smith @ruoccofabrizio

PR #1022, opened 2024-06-04
fix: Login to Azure in post provision step if not already

ross-p-smith:
Relates to #406 

If you run `azd auth login` from your devcontainer, you may not have ran `az login` which means that during the Post Provision hook, the `pfazure` command can fail

## How to Test
Dont login to Azure using `az login`
run `azd up` with PromptFlow enabled


adamdougal (approved):
LGTM!

Issue #979, opened 2024-05-24
Add a new Orchestrator "promptflow" that calls the prompt flow


tectonia:
Required by #406 

# Description

Create a prompt flow orchestrator. 

# Tasks

_To be filled in by the engineer picking up the subtask

- [ ] Task 1
- [ ] Task 2
- [ ] ...

PR #1026, opened 2024-06-08
feat: Add a new Orchestrator "prompt_flow"

tectonia:
This pull request introduces a new orchestration strategy called `Prompt Flow` in the codebase. The changes include adding new environment variables, updating the orchestration strategy enumeration, creating a new `PromptFlowOrchestrator` class, and updating the orchestrator selection logic to include the new strategy. Additionally, the `LLMHelper` class has been updated to support the new strategy, and unit tests have been added to verify the changes. Lastly, the local deployment documentation has been updated to reflect these changes.

### New Environment Variables:
* [`.env.sample`](diffhunk://#diff-088d9f35d23a4347d221d71dd49b02b95001dff4abe637a40fe0bc04d502049cL49-R54): Added new environment variables related to the Prompt Flow orchestration strategy.
* [`code/backend/batch/utilities/helpers/env_helper.py`](diffhunk://#diff-7503275e90a545dba3519622ff9d0866a73e814ddbfbb61156137721bf2a6552R230-R237): Updated the `__load_config` method to load the new environment variables.
* `infra/prompt-flow/create-prompt-flow.sh`: Updates the script to set environment variables locally and on web app. Assigns Azure ML Data Scientist role to web app for listing keys and accessing the scoring endpoint. 

### Orchestration Strategy Changes:
* [`code/backend/batch/utilities/orchestrator/orchestration_strategy.py`](diffhunk://#diff-17c4070b2508f248ea230788b6ba6bf5835c8409160151777dfa46daf74861dfR8): Added `PROMPT_FLOW` to the `OrchestrationStrategy` enumeration.
* [`code/backend/batch/utilities/orchestrator/strategies.py`](diffhunk://#diff-26dd9ea3cc5a023de6cb051014f4b3711a86a9d1a22b692b20b67da948d08568R15-R16): Updated the `get_orchestrator` method to return an instance of `PromptFlowOrchestrator` when the `PROMPT_FLOW` strategy is selected.

### New `PromptFlowOrchestrator` Class:
* [`code/backend/batch/utilities/orchestrator/prompt_flow.py`](diffhunk://#diff-b1a03b98fc199307899ee3bb4cb1a2ae51a1dc7e3d2af46f8aa3a68b2e8a8e82R1-R71): Created a new `PromptFlowOrchestrator` class that implements the orchestration logic for the Prompt Flow strategy.

### `LLMHelper` Class Changes:
* [`code/backend/batch/utilities/helpers/llm_helper.py`](diffhunk://#diff-43896759bad4bc2119807845cbb205a535d9caec19b671e07dcb6cfe73ec421aR9-R10): Updated the `LLMHelper` class to support the Prompt Flow strategy. This includes adding new methods to interact with the Azure ML workspace and transform the chat history. [[1]](diffhunk://#diff-43896759bad4bc2119807845cbb205a535d9caec19b671e07dcb6cfe73ec421aR9-R10) [[2]](diffhunk://#diff-43896759bad4bc2119807845cbb205a535d9caec19b671e07dcb6cfe73ec421aR159-R189)

### Unit Tests:
* [`code/tests/utilities/helpers/test_llm_helper.py`](diffhunk://#diff-65ac558c098ddef67bfe9c8f3b83a0a1616293900b3092778b55ceb0b1333863R16-R20): Added new unit tests to verify the changes in the `LLMHelper` class. [[1]](diffhunk://#diff-65ac558c098ddef67bfe9c8f3b83a0a1616293900b3092778b55ceb0b1333863R16-R20) [[2]](diffhunk://#diff-65ac558c098ddef67bfe9c8f3b83a0a1616293900b3092778b55ceb0b1333863R34-R38) [[3]](diffhunk://#diff-65ac558c098ddef67bfe9c8f3b83a0a1616293900b3092778b55ceb0b1333863R140-R233)
* [`code/tests/utilities/orchestrator/test_prompt_flow.py`](diffhunk://#diff-e736615ac30ba808134a821947447bd1887e40138f05e8b7c340f50ea55e2c24R1-R127): Added new unit tests to verify the `PromptFlowOrchestrator` class.

### Documentation Updates:
* `docs/LOCAL_DEPLOYMENT.md`, `docs/TEAMS_LOCAL_DEPLOYMENT.md`: Updated the local deployment documentation to reflect the new orchestration strategy and environment variables. [[1]](diffhunk://#diff-c79519d2609bc70c20c082a1fdbf47dfca79e43ec4342b8b9dba87f46be94096L212-R212) [[2]](diffhunk://#diff-d9eab1db2b9ede296201514861e1ef77aef18446fdf8db3e8b52a0f5b30b36a5L86-R86)

### Fixes local configuration of frontend
* `code/frontend/vite.config.ts`: Updates the vite configuration so port is exposed locally when running from a dev container. This fixes an issue some users where experiencing. 

## Limitations:
* Does not support advanced image processing, integrated vectorization etc.
* Has not been tested with rbac set as AZURE_AUTH_TYPE (more detail #979)

Closing #979


code/backend/batch/utilities/helpers/llm_helper.py
```diff
@@ -154,3 +156,34 @@ def get_sk_service_settings(self, service: AzureChatCompletion):
                 max_tokens=self.llm_max_tokens,
             ),
         )
+
+    def get_ml_client(self):
+        if not hasattr(self, "_ml_client"):
+            self._ml_client = MLClient(
+                DefaultAzureCredential(),
```
adamdougal, line 163: I think this is RBAC authentication, which isn't fully supported yet. Should we change this to keys instead?

I _think_ you can use `AzureKeyCredential` instead.

code/backend/batch/utilities/helpers/llm_helper.py
```diff
@@ -154,3 +156,34 @@ def get_sk_service_settings(self, service: AzureChatCompletion):
...
+                self.env_helper.AZURE_RESOURCE_GROUP,
+                self.env_helper.AZURE_ML_WORKSPACE_NAME,
+            )
+        return self._ml_client
+
+    def get_endpoint_name(self):
+        return self.env_helper.PROMPT_FLOW_ENDPOINT_NAME
+
+    def get_deployment_name(self):
+        return self.env_helper.PROMPT_FLOW_DEPLOYMENT_NAME
```
adamdougal, line 174: Not a big deal, but just wondering if the `PromptFlowOrchestrator` could have it's own reference to the `EnvHelper` instead of having to have pass through methods here?

code/backend/batch/utilities/helpers/llm_helper.py
```diff
@@ -154,3 +156,34 @@ def get_sk_service_settings(self, service: AzureChatCompletion):
...
+
+    def transform_chat_history_for_pf(self, chat_history):
+        transformed_chat_history = []
+        for i in range(0, len(chat_history), 2):
+            user_message = chat_history[i]['content']
+            assistant_message = chat_history[i+1]['content'] if i+1 < len(chat_history) else ''
```
adamdougal, line 180: There can be scenarios where there are multiple user messages before an assistant message. Usually when there is an error. Can we get the message based on the role field? Maybe we can just ignore any user messages that dont' have an assistant response?

code/backend/batch/utilities/helpers/llm_helper.py
```diff
@@ -154,3 +156,34 @@ def get_sk_service_settings(self, service: AzureChatCompletion):
...
+            )
+        return self._ml_client
+
+    def get_endpoint_name(self):
+        return self.env_helper.PROMPT_FLOW_ENDPOINT_NAME
+
+    def get_deployment_name(self):
+        return self.env_helper.PROMPT_FLOW_DEPLOYMENT_NAME
+
+    def transform_chat_history_for_pf(self, chat_history):
```
adamdougal, line 176: Just wondering if this should be a method within the `PromptFlowOrchestrator`? I think it'll only ever be used there.

code/backend/batch/utilities/orchestrator/prompt_flow.py
```diff
@@ -0,0 +1,71 @@
...
+            response = self.ml_client.online_endpoints.invoke(
+                endpoint_name=self.enpoint_name, request_file=file.name,
+                deployment_name=self.deployment_name
+            )
+            result = json.loads(response)
+            logger.debug(result)
+        except urllib.error.HTTPError as error:
+            logger.error("The request failed with status code: %s", str(error.code))
+            logger.error(error.info())
+            logger.error(error.read().decode("utf8", "ignore"))
```
adamdougal, line 52: Can we log this in one `logger.error`? Having them separate may make it harder to find all relevant information when something fails.

code/backend/batch/utilities/orchestrator/prompt_flow.py
```diff
@@ -0,0 +1,71 @@
...
+    def __init__(self) -> None:
+        super().__init__()
+        self.llm_helper = LLMHelper()
+
+        # Get the ML client and endpoint name
+        self.ml_client = self.llm_helper.get_ml_client()
+        self.enpoint_name = self.llm_helper.get_endpoint_name()
+        self.deployment_name = self.llm_helper.get_deployment_name()
+
+    async def orchestrate(
```
adamdougal, line 24: Could we maybe break this method into several smaller ones? It should make it easier to see the different steps involved, but also would allow us to remove some of the inline comments.

code/tests/utilities/orchestrator/test_prompt_flow.py
```diff
@@ -0,0 +1,127 @@
...
+
+        orchestrator.call_content_safety_input = MagicMock(return_value=None)
+        orchestrator.call_content_safety_output = MagicMock(return_value=None)
+
+        orchestrator.output_parser = OutputParserTool()
+
+        yield orchestrator
+
+def test_prompt_flow_init(orchestrator: PromptFlowOrchestrator):
+    assert orchestrator.ml_client is not None
```
adamdougal, line 42: Minor, but could this assert that it's the same as the mock you're assigning on line 17?

code/tests/utilities/orchestrator/test_prompt_flow.py
```diff
@@ -0,0 +1,127 @@
...
+    assert orchestrator.enpoint_name == "endpoint_name"
+    assert orchestrator.deployment_name == "deployment_name"
+
+@pytest.mark.asyncio
+async def test_content_safety_input(orchestrator: PromptFlowOrchestrator):
```
adamdougal, line 47: Can we name the tests based on the expected behaviour? E.g. something like `test_orchestrate_checks_question_for_content_safety` (just a suggestion, may be something better). That way, the tests become documentation for the code.

code/tests/utilities/orchestrator/test_prompt_flow.py
```diff
@@ -0,0 +1,127 @@
...
+    orchestrator.call_content_safety_output = MagicMock(return_value=None)
+    orchestrator.ml_client.online_endpoints.invoke = AsyncMock(return_value=chat_output)
+
+    # when
+    with patch('json.loads', return_value=chat_output):
+        response = await orchestrator.orchestrate(user_message, chat_history)
+
+    # then
+    orchestrator.call_content_safety_input.assert_called_once_with(user_message)
+    orchestrator.call_content_safety_output.assert_called_once_with(user_message, "answer")
```
adamdougal, line 100: Should these assertions move to the tests above and below?

code/tests/utilities/orchestrator/test_prompt_flow.py
```diff
@@ -0,0 +1,127 @@
...
+    )
+
+    # when
+    response = await orchestrator.orchestrate("bad question", [])
+
+    # then
+    assert response == content_safety_response
+
+@pytest.mark.asyncio
+async def test_orchestrate(orchestrator: PromptFlowOrchestrator):
```
adamdougal, line 72: Can we add a test for the error response too?

code/backend/batch/utilities/orchestrator/prompt_flow.py
adamdougal, line 52: Should we also throw a new exception with the http error? At the moment it will continue and then fail on line 57.

infra/prompt-flow/create-prompt-flow.sh
```diff
@@ -111,3 +116,16 @@ else
...
+az webapp config appsettings set --name $admin_webapp_name --resource-group $resource_group --settings PROMPT_FLOW_ENDPOINT_NAME=$endpoint_name PROMPT_FLOW_DEPLOYMENT_NAME=$deployment_name AZURE_ML_WORKSPACE_NAME=$aml_workspace
+
+echo "Assigning AzureML Data Scientist role to webapp"
+webapp_id=$(az webapp identity show --resource-group "$resource_group" --name $webapp_name --query principalId --output tsv)
+az role assignment create --role "AzureML Data Scientist" --scope "/subscriptions/${subscription_id}/resourceGroups/${resource_group}/providers/Microsoft.MachineLearningServices/workspaces/${aml_workspace}" --assignee-object-id "$webapp_id" --assignee-principal-type ServicePrincipal
+az role assignment create --role "AzureML Data Scientist" --scope "/subscriptions/${subscription_id}/resourceGroups/${resource_group}/providers/Microsoft.MachineLearningServices/workspaces/${aml_workspace}/onlineEndpoints/${endpoint_name}" --assignee-object-id "$webapp_id" --assignee-principal-type ServicePrincipal
```
adamdougal, line 131: Can all of the above be added to the bicep file instead? The web apps don't reload config, so would need to be restarted after this script runs. It may require us to move the endpoint and deployment names there too, and then pass them to this script instead, but that's okay I think.

adamdougal (changes requested):

code/backend/batch/utilities/orchestrator/prompt_flow.py
tectonia: done.

code/backend/batch/utilities/helpers/llm_helper.py
tectonia: done!

code/backend/batch/utilities/orchestrator/prompt_flow.py
tectonia: I popped out the data into file transformation to a separate method but thought the rest made sense to stay in orchestrate especially that it follows the same orchestrate format as lang chain, semantic kernel and open ai functions orchestrators. Happy to discuss and make further changes if you think this is still not very readable.

code/tests/utilities/orchestrator/test_prompt_flow.py
tectonia: done

code/backend/batch/utilities/orchestrator/prompt_flow.py
tectonia: good catch! added

code/tests/utilities/orchestrator/test_prompt_flow.py
tectonia: fixed!

code/backend/batch/utilities/helpers/llm_helper.py
tectonia: good point, fixed as suggested

code/backend/batch/utilities/orchestrator/prompt_flow.py
```diff
@@ -0,0 +1,103 @@
...
+    def transform_chat_history(self, chat_history):
+        # Transform chat history into a format that the Prompt Flow service can understand
+        transformed_chat_history = []
+        i = 0
+        while i < len(chat_history):
+            if chat_history[i]['role'] == 'user':
+                user_message = chat_history[i]['content']
+                i += 1
+                assistant_message = ''
+                # Find the next assistant message
```
adamdougal, line 78: I think we want to find the latest user message, followed by the assistant message. E.g. If we have:
```
user: Hello
user: What year is it?
assistant: 2024
```

We want the final two messages, not the first and last.

Can we also add a test to cover this scenario?

infra/prompt-flow/create-prompt-flow.sh
tectonia: as discussed in offline meeting, unfortunately no - as endpoint will not exist before this script is run.

code/backend/batch/utilities/orchestrator/prompt_flow.py
tectonia: after discussion offline, changed this to have empty string for output following user message with no assistant response.

code/backend/batch/utilities/orchestrator/prompt_flow.py
tectonia: broke out the transform_data_into_file method. rest left as is as this is the same structure followed by lang chain, semantic kernel and open ai functions orchestrators.

code/backend/batch/utilities/helpers/llm_helper.py
tectonia: discussed offline. ml client does not support keys as aml workspaces do not use keys. using key to auth to endpoint would require changing the invoke method to generic python sdk and getting endpoint key in script and uploading that to key vault. decided to keep the current auth method as hybrid of keys and rbac (using rbac to list keys).

adamdougal (approved):
LGTM! Great work!

We need to make some changes to support citations but we can add that later.

adamdougal (approved):

PR #1089, opened 2024-06-25
feat: Include citations in response when using prompt flow

adamdougal:
- This enabled me to remove a bunch of citation parsing from the flow
- This also enabled me to remove the citation output files
- Added verbose output for tests to aid debugging
- Fix issue with pfazure not being installed

Required by https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/issues/406


Makefile
```diff
@@ -36,7 +36,7 @@ python-test: ## 🧪 Run Python unit + functional tests
 
 unittest: ## 🧪 Run the unit tests
 	@echo -e "\e[34m$@\e[0m" || true
-	@poetry run pytest -m "not azure and not functional" $(optional_args)
+	@poetry run pytest -vvv -m "not azure and not functional" $(optional_args)
```
ross-p-smith, line 39: Curious - what does `-vvv` do? Couldn't find it in the docs

adamdougal: It makes the test output more verbose. I was struggling to debug a failing test, adding this shows the full diff rather then a snippet. 

https://docs.pytest.org/en/7.1.x/how-to/output.html#verbosity
> Using higher verbosity levels (-vvv, -vvvv, …) is supported, but has no effect in pytest itself at the moment, however some plugins might make use of higher verbosity.

I'm not sure if any plugins need the 3 v's, but I added it just in case!

ross-p-smith (approved):
LovinIt

PR #1105, opened 2024-07-02
fix: Smooth Error Handling for Subscription Token Exhaustion

Pavan-Microsoft:
## Purpose
Smooth Error Handling for Subscription Token Exhaustion

- [ ] Yes
- [x] No

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
npm run dev
```

## What to Check
Use the platform heavily until the subscription token limit is reached.
Observe the alert popup.
Verify that the message.
after

### DIFF
--- a/code/backend/constants.py
+++ b/code/backend/constants.py
@@ -1,2 +1,0 @@
-ERROR_429_MESSAGE = "We're currently experiencing a high number of requests for the service you're trying to access. Please wait a moment and try again.",
-ERROR_GENERIC_MESSAGE = "An error occurred. Please try again. If the problem persists, please contact the site administrator."
--- a/code/backend/constants.py
+++ b/code/backend/constants.py
@@ -1,0 +1,2 @@
+ERROR_429_MESSAGE = "We're currently experiencing a high number of requests for the service you're trying to access. Please wait a moment and try again.",
+ERROR_GENERIC_MESSAGE = "An error occurred. Please try again. If the problem persists, please contact the site administrator."
--- a/code/create_app.py
+++ b/code/create_app.py
@@ -20,7 +20,7 @@ from backend.batch.utilities.helpers.config.conversation_flow import Conversatio
 from azure.mgmt.cognitiveservices import CognitiveServicesManagementClient
 from azure.identity import DefaultAzureCredential
 
-ERROR_429_MESSAGE = "We're currently experiencing a high number of requests for the service you're trying to access. Please wait a moment and try again.",
+ERROR_429_MESSAGE = "We're currently experiencing a high number of requests for the service you're trying to access. Please wait a moment and try again."
 ERROR_GENERIC_MESSAGE = "An error occurred. Please try again. If the problem persists, please contact the site administrator."
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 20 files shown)


code/backend/constants.py
```diff
@@ -0,0 +1,2 @@
+ERROR_429_MESSAGE = "We're currently experiencing a high number of requests for the service you're trying to access. Please wait a moment and try again.",
```
adamdougal, line 1: Instead of having these constants in a separate file, can they live at the top of the `create_app.py`? Keeps the context together that way.

code/create_app.py
```diff
@@ -346,14 +347,11 @@ def conversation_azure_byod():
...
             logger.exception("Exception in /api/conversation | %s", error_message)
-            return (
-                jsonify(
-                    {
-                        "error": "Exception in /api/conversation. See log for more details."
-                    }
-                ),
-                500,
-            )
+            if hasattr(e, 'code'):
```
adamdougal, line 350: Instead of having to check for an attribute, do we know if a specific type of exception is thrown that would allow us to have another `except` block?

code/frontend/src/pages/chat/Chat.tsx
```diff
@@ -114,10 +114,13 @@ const Chat = () => {
...
-        console.error(result);
-        alert(
-          "An error occurred. Please try again. If the problem persists, please contact the site administrator."
-        );
+        if (e instanceof Error) {
+          alert(e.message);
+        }
+        else
+        {
+          alert('An unknown error occurred');
```
adamdougal, line 122: Can we keep this as
```suggestion
          alert('An error occurred. Please try again. If the problem persists, please contact the site administrator.');
```

adamdougal (changes requested):
This is a great addition!

code/backend/constants.py
Roopan-Microsoft: Yeah, we can do that @adamdougal. Thought to maintain in a single place in case if we get more constants in future.

code/frontend/src/pages/chat/Chat.tsx
Pavan-Microsoft: Updated error message to: 'An error occurred. Please try again. If the problem persists, please contact the site administrator.'

code/create_app.py
Pavan-Microsoft: Added specific exception type for 'RateLimitError'. This change eliminates the need to check for an attribute by allowing us to use an additional except block for this specific exception type.

code/backend/constants.py
Pavan-Microsoft: Moved constants to the top of create_app.py

code/tests/test_app.py
```diff
@@ -688,7 +688,7 @@ def test_conversation_azure_byod_returns_500_when_exception_occurs(
         # then
         assert response.status_code == 500
         assert response.json == {
-            "error": "Exception in /api/conversation. See log for more details."
+            "error": "An error occurred. Please try again. If the problem persists, please contact the site administrator."
         }
```
adamdougal, line 693: Can we add a unit test for the rate limit exception?

adamdougal (commented):
Thanks for making those changes! I think all we're missing now is additional unit tests for this exception being thrown.

code/tests/test_app.py
Pavan-Microsoft: Yes, we've added a unit test for the rate limit exception now.

adamdougal (approved):
Thanks for this ❤️

PR #1108, opened 2024-07-03
ci: Enable `report-only-changed-files` in `comment_coverage.yml`

cecheta:
## Purpose

* The bug that prevented `report-only-changed-files` working on forks has now been fixed, therefore setting the value to `true`

- [ ] Yes
- [x] No


adamdougal (approved):
Nice work getting that fixed!

PR #1110, opened 2024-07-04
fix: exception occurred deleting files

Prashant-Microsoft:
## Purpose
* Exception occurred deleting files

- [ ] Yes
- [x] No

## How to Test
*  In the admin site if you go to delete data section, then you will get the exception

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code
```
Launch admin site
```

## What to Check
- Launch the admin site and navigate to delete data section

### DIFF
--- a/code/backend/batch/utilities/search/integrated_vectorization_search_handler.py
+++ b/code/backend/batch/utilities/search/integrated_vectorization_search_handler.py
@@ -48,12 +48,13 @@ class IntegratedVectorizationSearchHandler(SearchHandlerBase):
     def output_results(self, results):
         files = {}
         for result in results:
-            id = result["chunk_id"]
-            filename = result["title"]
-            if filename in files:
-                files[filename].append(id)
-            else:
-                files[filename] = [id]
+            if result["chunk_id"] is not None and result["title"] is not None:
+                id = result["chunk_id"]
+                filename = result["title"]
+                if filename in files:
+                    files[filename].append(id)
+                else:
+                    files[filename] = [id]
 
         return files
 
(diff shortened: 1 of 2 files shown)


code/backend/batch/utilities/search/integrated_vectorization_search_handler.py
```diff
@@ -48,12 +48,13 @@ def get_files(self):
     def output_results(self, results):
         files = {}
         for result in results:
-            id = result["chunk_id"]
-            filename = result["title"]
-            if filename in files:
-                files[filename].append(id)
-            else:
-                files[filename] = [id]
+            if result["chunk_id"] is not None and result["title"] is not None:
```
adamdougal, line 51: Can we add a unit test for this case?

adamdougal (changes requested):
Do we know under what circumstances this can happen? Do you have a copy or a screenshot of the error? Thanks

Prashant-Microsoft:
> Do we know under what circumstances this can happen? Do you have a copy or a screenshot of the error? Thanks

In some of the documents we are getting title field as null, that case it will give us error
image

adamdougal:
Gotcha! My concern with this fix is that, yes it stops the exception, but it also stops the file being outputted, which means it cannot be deleted.

Could we address the issue of the title not being there instead?

Roopan-Microsoft:
> Gotcha! My concern with this fix is that, yes it stops the exception, but it also stops the file being outputted, which means it cannot be deleted.
> 
> Could we address the issue of the title not being there instead?

Yes @adamdougal - We are also trying the same scenario to repro that issue. Will figure it out

Prashant-Microsoft:
> Gotcha! My concern with this fix is that, yes it stops the exception, but it also stops the file being outputted, which means it cannot be deleted.
> 
> Could we address the issue of the title not being there instead?

@adamdougal Seems like this issue is intermittent, now we are not able to get the title field as null, everything working properly but still, we are working on it to repro the issue

Prashant-Microsoft:
> Gotcha! My concern with this fix is that, yes it stops the exception, but it also stops the file being outputted, which means it cannot be deleted.
> 
> Could we address the issue of the title not being there instead?

@adamdougal Seems like this issue is intermittent, now we are not able to get the title field as null, everything working properly but still, we are working on it to repro the issue

adamdougal:
Closing as we can't reproduce the error.

PR #1111, opened 2024-07-05
fix: Handle RateLimitError in BadRequestException

Pavan-Microsoft:
## Purpose
Handle rate limit errors occurring within BadRequestException. In some cases, rate limit errors are not captured by the OpenAI RateLimitError exception type. Instead, they appear in the inner message of BadRequestException. These changes address this issue.
BeforeFix4410_1

- [ ] Yes
- [ x] No

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

## What to Check
Use the platform heavily until the subscription token limit is reached.
Observe the alert popup.
Verify that the message.
AfterFix4410


adamdougal:
Based on your screenshot, the code that was being used when you saw this was before we started catching the `RateLimitError` exception which was added by https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/1105

Are you sure this is still required?

Pavan-Microsoft:
> Based on your screenshot, the code that was being used when you saw this was before we started catching the `RateLimitError` exception which was added by #1105
> 
> Are you sure this is still required?

Yes, it is still required. We have observed that the rate limit error can also occur within the BadRequest exception type, which means it is not always caught by the OpenAI.RateLimitError exception type alone.

code/create_app.py
```diff
@@ -24,6 +24,24 @@
 ERROR_GENERIC_MESSAGE = "An error occurred. Please try again. If the problem persists, please contact the site administrator."
 logger = logging.getLogger(__name__)
 
+# Helper function to extract inner error code safely
+def get_inner_error_code(e):
```
adamdougal, line 28: This didn't work for me. It caused a subsequent exception:
```
openai.BadRequestError: Error code: 400 - {'error': {'requestid': '89e0d2f5-03c3-4366-ad77-fd80d37ed3ae', 'code': 400, 'message': "An error occurred when calling Azure OpenAI: Server responded with status 429. Error message: {'error': {'code': '429', 'message': 'Rate limit is exceeded. Try again in 10 seconds.'}}"}}
ERROR:create_app:Exception in /api/conversation | Expecting value: line 1 column 1 (char 0)
Traceback (most recent call last):
  File "/workspaces/chat-with-your-data-solution-accelerator/code/create_app.py", line 39, in get_inner_error_code
    inner_error_json = json.loads(inner_error_json_part)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

adamdougal (commented):
It's a shame it's not easy to extract the 429 error from the response/exception! I'm wondering if we take a slightly different approach with this to keep it simple. Instead of catching `RateLimitError`, we could catch it's parent class `ApiStatusError`. We can then do something like this:
```python
        except APIStatusError as e:
            error_message = str(e)
            logger.exception("Exception in /api/conversation | %s", error_message)
            response_json = e.response.json()
            response_message = response_json.get("error", {}).get("message", "")
            if "Rate limit is exceeded" in response_message:
                return jsonify({"error": ERROR_429_MESSAGE}), 429
            
            return jsonify({"error": ERROR_GENERIC_MESSAGE}), 500
```
This would then allow us to remove all of the other checks we have in the other catch blocks. What do you think?

Pavan-Microsoft:
> It's a shame it's not easy to extract the 429 error from the response/exception! I'm wondering if we take a slightly different approach with this to keep it simple. Instead of catching `RateLimitError`, we could catch it's parent class `ApiStatusError`. We can then do something like this:
> 
> ```python
>         except APIStatusError as e:
>             error_message = str(e)
>             logger.exception("Exception in /api/conversation | %s", error_message)
>             response_json = e.response.json()
>             response_message = response_json.get("error", {}).get("message", "")
>             if "Rate limit is exceeded" in response_message:
>                 return jsonify({"error": ERROR_429_MESSAGE}), 429
>             
>             return jsonify({"error": ERROR_GENERIC_MESSAGE}), 500
> ```
> 
> This would then allow us to remove all of the other checks we have in the other catch blocks. What do you think?



> ```python
> APIStatusError 
> ```

As discussed, We have updated the code to handle ratelimit exceptions using the APIStatusError type. Additionally, we catch any other exceptions with the generic Exception type. (as we've noticed that ValueError exceptions are not being caught with APIStatusError).

adamdougal:
@Pavan-Microsoft Changes look good! Can we also apply this to the `/conversation/custom` endpoint for consistency? Thanks!

Pavan-Microsoft:
> @Pavan-Microsoft Changes look good! Can we also apply this to the `/conversation/custom` endpoint for consistency? Thanks!

Updated the rate limit exception handling for **/conversation/custom** to be consistent with the **BYOD** flow. Additionally, we added a **code** check because we observed that the error message in the **/conversation/custom** flow does not include the '**429**' error number.
custom_ratelimi

adamdougal (approved):
This is great!

PR #1128, opened 2024-07-11
fix: Fixed Admin Interface: Explore Data-Files visibility issue

Prasanjeet-Microsoft:
## Purpose
Fixed files visibility issue under explore data section in Admin interface where end user was not able to see more than 10 files and after this code fix we are able to see all the files.

- [ ] Yes
- [x] No

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

## What to Check
Upload more than 10 files from ingest data section in admin interface and verify that we are able to see all the files in explore data section. Please find below screenshots for your reference.

image

image

### DIFF
--- a/code/backend/batch/utilities/search/search_handler_base.py
+++ b/code/backend/batch/utilities/search/search_handler_base.py
@@ -12,12 +12,12 @@ class SearchHandlerBase(ABC):
         self.env_helper = env_helper
         self.search_client = self.create_search_client()
 
-    def search_with_facets(self, query: str, facets: list[str], facet_count: int = None):
+    def search_with_facets(self, query: str, facets: list[str], facet_count: int):
         if self.search_client is None:
             return None
 
         # Construct facet parameter based on facet_count
-        facet_param = facets[0] if facet_count is None else f"{facets[0]},count:{facet_count}"
+        facet_param = f"{facets[0]},count:{facet_count}"
 
         # Perform search with facets and facet_param
         return self.search_client.search(query, facets=[facet_param])
--- a/code/backend/batch/utilities/search/search_handler_base.py
+++ b/code/backend/batch/utilities/search/search_handler_base.py
@@ -12,12 +12,12 @@ class SearchHandlerBase(ABC):
         self.env_helper = env_helper
         self.search_client = self.create_search_client()
 
-    def search_with_facets(self, query: str, facets: list[str], facet_count: int):
+    def search_with_facets(self, query: str, facet: str, facet_count: int):
         if self.search_client is None:
             return None
 
         # Construct facet parameter based on facet_count
-        facet_param = f"{facets[0]},count:{facet_count}"
+        facet_param = f"{facet},count:{facet_count}"
 
         # Perform search with facets and facet_param
         return self.search_client.search(query, facets=[facet_param])
--- a/code/backend/batch/utilities/search/search_handler_base.py
+++ b/code/backend/batch/utilities/search/search_handler_base.py
@@ -12,21 +12,20 @@ class SearchHandlerBase(ABC):
         self.env_helper = env_helper
         self.search_client = self.create_search_client()
 
-    def search_with_facets(self, query: str, facets: list[str]):
+    def search_with_facets(self, query: str, facets: list[str], facet_count: int = None):
         if self.search_client is None:
             return None
-        return self.search_client.search(query, facets=facets)
 
-    def get_unique_files(self, search_text, field_name):
-        unique_files = set()  # Using a set to automatically handle uniqueness
-        results = self.search_client.search(search_text=search_text)
+        # Construct facet parameter based on facet_count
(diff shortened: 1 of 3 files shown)


code/backend/batch/utilities/search/search_handler_base.py
```diff
@@ -17,10 +17,16 @@ def search_with_facets(self, query: str, facets: list[str]):
             return None
         return self.search_client.search(query, facets=facets)
 
-    def get_unique_files(self, results, facet_key: str):
-        if results:
-            return [facet["value"] for facet in results.get_facets()[facet_key]]
-        return []
+    def get_unique_files(self, search_text, field_name):
+        unique_files = set()  # Using a set to automatically handle uniqueness
```
adamdougal, line 21: Can we use a unit test to enforce this rather than a comment?

code/backend/batch/utilities/search/search_handler_base.py
```diff
@@ -17,10 +17,16 @@ def search_with_facets(self, query: str, facets: list[str]):
...
+        results = self.search_client.search(search_text=search_text)
```
adamdougal, line 22: Didn't using facets handle the uniqness before?

adamdougal (commented):
Do we know why this caused the bug in the first place? Thanks

Prasanjeet-Microsoft:
> Do we know why this caused the bug in the first place? Thanks

Because in the admin interface under explore data section, we were not able to see more than 10 files earlier.

code/backend/batch/utilities/search/search_handler_base.py
Prasanjeet-Microsoft: No, problem was not with the uniqueness. Problem was with the facets count which has a default limit of the top 10 facets. Now we have added one extra parameter for facet count in search_with_facets function which will help us to retrieve all the documents from search client.

code/backend/batch/utilities/search/search_handler_base.py
```diff
@@ -12,10 +12,15 @@ def __init__(self, env_helper: EnvHelper):
...
         self.search_client = self.create_search_client()
 
-    def search_with_facets(self, query: str, facets: list[str]):
+    def search_with_facets(self, query: str, facets: list[str], facet_count: int = None):
         if self.search_client is None:
             return None
-        return self.search_client.search(query, facets=facets)
+
+        # Construct facet parameter based on facet_count
+        facet_param = facets[0] if facet_count is None else f"{facets[0]},count:{facet_count}"
```
adamdougal, line 20: I think we can hard code this to `f"{facets[0]},count:0"` to always return **all** results. I don't think we have a usecase of this function which needs to limit or will page over the results.

Can we also add some unit tests around this change? Thanks

Prasanjeet-Microsoft: Sure. We will remove if else condition and will directly pass **f"{facets[0]},count:0"** to facet_params. 

We haven't found any test cases for search_handler_base.py file.

adamdougal: Ah we should add it to one of the implementation classes tests file e.g. https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/main/code/tests/search_utilities/test_azure_search_handler.py

code/backend/batch/utilities/search/search_handler_base.py
Prasanjeet-Microsoft: Sure. Added test cases, can you please review it once

code/backend/batch/utilities/search/search_handler_base.py
Prasanjeet-Microsoft: Added test cases, can you please review it once.

code/backend/batch/utilities/search/search_handler_base.py
```diff
@@ -12,10 +12,15 @@ def __init__(self, env_helper: EnvHelper):
...
         self.search_client = self.create_search_client()
 
-    def search_with_facets(self, query: str, facets: list[str]):
+    def search_with_facets(self, query: str, facets: list[str], facet_count: int):
         if self.search_client is None:
             return None
-        return self.search_client.search(query, facets=facets)
+
+        # Construct facet parameter based on facet_count
+        facet_param = f"{facets[0]},count:{facet_count}"
```
adamdougal, line 20: Sorry I should have picked up on this sooner. 

`facets` is a list, but we're only getting the first one, meaning any extra are ignored. However, I don't think we actually ever pass in more than one, so I'm happy for the fix to either be:
- Include all `facets` in the `facets_param`
- Change the `facets` argument to be `facet: str`

Prasanjeet-Microsoft: @adamdougal We have made the changes. can you please review and approve.

adamdougal (approved):
LGTM!

PR #1131, opened 2024-07-12
fix: Disable send button until speech recording stopped

Rohini-Microsoft:
## Purpose
mic recording and send button functionality not in sync

Overlapping of "Listening.." and "Stop Generating" messages
After recording query when we send the it,  "Listening.." and "Stop Generating" are overlapping 

- [ ] Yes
- [x] No

## How to Test
once clicked on send button 
microphone recording should be stopped 
Overlapping of "Listening.." and "Stop Generating" messages
After recording query when we send the it,  "Listening.." and "Stop Generating" are overlapping 

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

## What to Check
1. click on mic to start the recording 
2. once we click on send button the recorded sentence is sent, but mice don't stop recording
3. if user keeps talking boat concatenate the first said sentence and currently speaking statement and combinedly provides the result. 

image

image

### DIFF
--- a/code/frontend/src/components/QuestionInput/QuestionInput.module.css
+++ b/code/frontend/src/components/QuestionInput/QuestionInput.module.css
@@ -35,6 +35,13 @@
     height: 23px;
 }
 
+.SendButtonDisabled{
+    pointer-events: none;
+    width: 24px;
+    height: 23px;
+    margin-right: 12px;
+}
+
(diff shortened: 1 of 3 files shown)


adamdougal (approved):
Nice! LGTM!

PR #1173, opened 2024-07-26
fix: Add fallback message for preview display issues with tables and special formatting

Pavan-Microsoft:
## Purpose
This PR introduces a fallback mechanism to enhance the user experience when the preview functionality fails to properly display tables, images, or other special formatting. The added message: "Tables, images, and other special formatting not shown in this preview. Please follow the link to review the original document." will inform users of these limitations and guide them to the original document for accurate review. 

- [ ] Yes
- [x ] No

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

## What to Check
Verify that the following:
fallbackmessage

### DIFF
--- a/code/backend/batch/utilities/document_loading/word_document.py
+++ b/code/backend/batch/utilities/document_loading/word_document.py
@@ -24,10 +24,10 @@ class WordDocumentLoading(DocumentLoadingBase):
         return file
 
     def _get_opening_tag(self, heading_level: int) -> str:
-        return f"<{self.doc_headings_to_markdown_tags.get(f'{heading_level}', '')}>"
+        return f"<{self.doc_headings_to_markdown_tags.get(f'{heading_level}', 'p')}>"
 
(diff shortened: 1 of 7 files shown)


adamdougal:
I'm wondering if we can just put this in the front end code, between these two lines https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/main/code/frontend/src/pages/chat/Chat.tsx#L383-L384 rather than putting front end html in the backend code. What do you think?

Pavan-Microsoft:
> I'm wondering if we can just put this in the front end code, between these two lines https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/main/code/frontend/src/pages/chat/Chat.tsx#L383-L384 rather than putting front end html in the backend code. What do you think?

Hi @adamdougal Moved code to frontend.

adamdougal (approved):

PR #1200, opened 2024-08-01
fix: Handled the file format if its in UPPER CASE

Roopan-Microsoft:
## Purpose

* If user uploads any file with uppercase file format, It was not processing. So we are converting into lower case and processing now.

- [ ] Yes
- [x ] No

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* Validate by uploading any file name with upper case extension


code/backend/batch/utilities/helpers/embedders/push_embedder.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, blob_client: AzureBlobStorageClient, env_helper: EnvHelper):
             self.embedding_configs[ext] = processor
 
     def embed_file(self, source_url: str, file_name: str):
-        file_extension = file_name.split(".")[-1]
+        file_extension = file_name.split(".")[-1].lower()
```
adamdougal, line 41: Can we add/modify a unit test to ensure this is present? Could just uppercase one of the test file extentions https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/blob/main/code/tests/utilities/helpers/test_push_embedder.py#L129C1-L130C1

adamdougal (approved):

PR #1219, opened 2024-08-08
fix: Updated the column name as "id" for the azureSearchFieldId column in bicep code

Roopan-Microsoft:
## Purpose

* Updated the column name as "id" for the azureSearchFieldId column in bicep code

- [ ] Yes
- [ x] No

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...


Prasanjeet-Microsoft (approved):
Approved.

Issue #251, opened 2024-01-31
Unable to modify index search field names


trevor-mcgirr:
The issue is related to using the automated Microsoft “Indexer” instead of ingesting the documents through the Admin UI in this application. When setting the field names of the new Index we need to modify some of the existing field  names, but this seems to cause a mapping issue with the application. Despite updating the searchable names to match the index names, there seems to be a consistent error relating to needing “content_vector” or another field “chunk”.  Are certain fields required for Azure Search for Langchain integration or are there considerations to make when changing index field names and mapping to the application? 

In short: How can we change the application to search an index using custom field names?

### This issue is for a: 
```
- [x ] bug report 
- [ ] feature request
- [ ] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

### Minimal steps to reproduce
> Create an index using an indexer with unique field names and attempt to query the new index.

Modified Index (Custom Index)
<img width="1086" alt="Screenshot 2024-01-31 at 11 04 39 AM" src="https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/assets/155759150/b5021311-9bcb-4fc2-a546-20fcbde6c6e3">

Updated Searchable Fields (AzureSearchHelper.py)
<img width="707" alt="Screenshot 2024-01-31 at 11 09 14 AM" src="https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/assets/155759150/d5e9d300-85bd-4448-9f50-1f35943a98a0">

ENV variables (Set/Default if not defined)
<img width="394" alt="Screenshot 2024-01-31 at 11 15 26 AM" src="https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/assets/155759150/93ea8d8a-5513-4e21-a108-c65197774971">

### Any log messages given by the failure
Similar errors for “content_vector” and “chunk” based on various ENV variables attempted

>azure.core.exceptions.HttpResponseError: () The request is invalid. Details: The property 'chunk' does not exist on type 'search.documentFields'. Make sure to only use property names that are defined by the type.
Code: 
Message: The request is invalid. Details: The property 'chunk' does not exist on type 'search.documentFields'. Make sure to only use property names that are defined by the type.
Exception in /api/conversation/custom

>{
[...]

PR #1220, opened 2024-08-08
fix: Updated the column name as "id" for the azureSearchFieldId column in main.json file

Prasanjeet-Microsoft:
## Purpose

* Updated the column name as "id" for the azureSearchFieldId column in main.json file

- [ ] Yes
- [x] No

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

## What to Check
Check whether the value of AZURE_SEARCH_FIELDS_ID variable got updated to "id" in search index after doing the deployment. Please find below screenshot for your reference:

image

### DIFF
--- a/infra/main.json
+++ b/infra/main.json
@@ -136,7 +136,7 @@
     },
     "azureSearchFieldId": {
       "type": "string",
-      "defaultValue": "Id",
+      "defaultValue": "id",
       "metadata": {
         "description": "Id columns"
       }


ross-p-smith:
Hi @Prasanjeet-Microsoft - why is this change necessary? Could be perfectly valid, would be great to understand the need though so it can be tracked.

Prasanjeet-Microsoft:
> Hi @Prasanjeet-Microsoft - why is this change necessary? Could be perfectly valid, would be great to understand the need though so it can be tracked.

This changes was made as adam has asked to make **index search fields name** configurable via env vars to save the user needing to modify code. But the id field was not updated properly in main.json file so we have updated it now. Please find below github issue link for the same.

https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/issues/251

Thank you.

Roopan-Microsoft (approved):

PR #1290, opened 2024-09-06
feat: Automation flow for one click deployment

Prashant-Microsoft:
## Purpose
Automation for one click deployment and cleanup the resources

- [ ] Yes
- [x] No

## How to Test
*  Run the automation pipeline

### DIFF
--- a/.github/workflows/az-deploy.yml
+++ b/.github/workflows/az-deploy.yml
@@ -3,7 +3,7 @@ name: Deploy Azure Resources
 on:
   push:
     branches:
-      - PSL-Automation-7724
+      - main
 
 jobs:
   deploy:
--- a/.github/workflows/az-deploy.yml
+++ b/.github/workflows/az-deploy.yml
@@ -34,4 +34,4 @@ jobs:
             --name automation-CWYD \
             --location eastus \
             --template-file infra/main.bicep \
-            --parameters environmentName=pslautomation2 location=eastus2
+            --parameters environmentName=pslautomationtest location=eastus2
--- a/.github/workflows/az-deploy.yml
+++ b/.github/workflows/az-deploy.yml
@@ -1,0 +1,37 @@
+name: Deploy Azure Resources
+
+on:
+  push:
+    branches:
+      - PSL-Automation-7724
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v3
+
+      - name: Setup Azure CLI
+        run: |
+          curl -sL https://aka.ms/InstallAzureCLIDeb | sudo bash
+          az --version  # Verify installation
+
+      - name: Login to Azure
+        run: |
+          az login --service-principal -u ${{ secrets.AUTO_AZURE_CLIENT_ID }} -p ${{ secrets.AUTO_AZURE_CLIENT_SECRET }} --tenant ${{ secrets.AUTO_AZURE_TENANT_ID }}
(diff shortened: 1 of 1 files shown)


.github/workflows/az-deploy.yml
```diff
@@ -0,0 +1,37 @@
...
+        run: az bicep install
+
+      - name: Deploy Bicep Template
+        id: deploy
+        run: |
+          set -e
+          az deployment sub create \
+            --name automation-CWYD \
+            --location eastus \
+            --template-file infra/main.bicep \
```
adamdougal, line 36: If this is to test the one click deploy button then I think this should use the `infra/main.json` file instead of the bicep file.

.github/workflows/az-deploy.yml
```diff
@@ -0,0 +1,37 @@
...
+            --parameters environmentName=pslautomationtest location=eastus2
```
adamdougal, line 37: Should there also be a step to tear this down and clean up the resources?

Prashant-Microsoft: yes, we are implementing that also.

.github/workflows/az-deploy.yml
Prashant-Microsoft: There are basically two ways:
1. we can use directly bicep files to deploy the resource
2.  or else first we need to generate arm template (nothing but main.json file) from bicep file. after it we can deploy the resources.
So, we are using step 1, instead of creating arm templates, we directly using bicep file

Is there any specific reason that you are asking to use main.json instead of bicep module? Please let us know.

.github/workflows/az-deploy.yml
```diff
@@ -0,0 +1,37 @@
...
+        uses: actions/checkout@v3
+
+      - name: Setup Azure CLI
+        run: |
+          curl -sL https://aka.ms/InstallAzureCLIDeb | sudo bash
+          az --version  # Verify installation
+
+      - name: Login to Azure
+        run: |
+          az login --service-principal -u ${{ secrets.AUTO_AZURE_CLIENT_ID }} -p ${{ secrets.AUTO_AZURE_CLIENT_SECRET }} --tenant ${{ secrets.AUTO_AZURE_TENANT_ID }}
```
ross-p-smith, line 23: We've created new variables `AUTO_` - and not re-using the existing ones - we should update the existing github variables to your new subscription values and use them. At the moment, this strikes me as making things more complicated

ross-p-smith:
hi there - can you help me understand what this PR is trying to do? What is the end goal? There is a ci.yml workflow which calls `make deploy` that deploys the infrastructure. I'm happy to jump on a call and discuss - but I feel you are re-inventing the wheel.

.github/workflows/az-deploy.yml
Prashant-Microsoft: For testing purposes, we have created these variables so that they won't affect the current deployment. In the end, we will be using the same variables.

ross-p-smith:
I've just looked into this and the credentials used to deploy have expired. I feel that it is best that you enter your credentials into the existing GH actions variables and get the build working again

.github/workflows/az-deploy.yml
ross-p-smith: The current deployment is not working as the spn has expired - you are free to re-use those variables if you like

ross-p-smith:
I feel that you have fixed the current build and that it is deploying to your subscription and destroying on every push to `main` - I feel that you should close this PR now as it adds no "new" benefit

PR #1323, opened 2024-09-16
feat: Upload to AI Search in Batches

### LINKED ISSUE #1267: Batch calls to AI Search to Support large PDFs
# Motivation

Trying to index a couple PDFs that are individually >18 Megs in size with >800 pages each.  We're getting a 413 request too big error from AI Search.  (That's after increasing function timeout and increasing token allocation to embedding to get to this error.) From what I can tell the pages are extracted, chunked, and embedded, but the code then tries to upload all chunks to the AI search in one call.  Would be great if this could be broken up into batches.

# Requirements

- Send documents to AI Search in batches allowing large PDF's to be indexed. 

# Tasks

- [ ] Task 1
- [ ] Task 2
- [ ] ...

cmw2:
## Purpose
* Allows large PDFs to be ingested.
* Fixes #1267 

- [ ] Yes
- [x] No

## How to Test
*  Get the code
*  Follow steps document in Local Deployment

* Test the code
  * Ingest a large file.  My tested one had ~850 pages and was 18.5 Megs in size.

## What to Check
* All chunks should be in search index (can take a while)

## Other Information

* Modified existing upload test to pass with 1 batch.
* Added another test to check for multiple batches.
* Also had to do some code cleanup per black and flake8, though they conflicted in some rules so added exceptions in flake config file.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -2,3 +2,4 @@
 max-line-length = 88
 extend-ignore = E501
 exclude = .venv
+ignore = E203, W503
--- a/code/backend/batch/utilities/helpers/embedders/push_embedder.py
+++ b/code/backend/batch/utilities/helpers/embedders/push_embedder.py
@@ -79,12 +79,18 @@ class PushEmbedder(EmbedderBase):
             for document in documents:
                 documents_to_upload.append(self.__convert_to_search_document(document))
 
-        response = self.azure_search_helper.get_search_client().upload_documents(
-            documents_to_upload
-        )
-        if not all([r.succeeded for r in response]):
-            logger.error("Failed to upload documents to search index")
-            raise Exception(response)
+        # Upload documents (which are chunks) to search index in batches
+        if documents_to_upload:
+            batch_size = self.env_helper.AZURE_SEARCH_DOC_UPLOAD_BATCH_SIZE
+            search_client = self.azure_search_helper.get_search_client()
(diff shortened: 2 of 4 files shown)


.flake8
```diff
@@ -2,3 +2,4 @@
 max-line-length = 88
 extend-ignore = E501
 exclude = .venv
+ignore = E203, W503
```
adamdougal, line 5: Heya! Can you share the error you were getting with this? I can't seem to reproduce it. Thanks

cmw2: If I take out the line and run flake8 at the root level I get:
`./code/backend/batch/utilities/helpers/embedders/push_embedder.py:87:46: E203 whitespace before ':'`

Then if I add the ignore for E203 I get many errors like:
`./code/backend/batch/utilities/common/answer.py:27:13: W503 line break before binary operator`

I don't get functional errors, just in flake, but am not able to commit with those errors.  And if I try to adjust the code to pass flake, black reformats it back during the commit.

adamdougal: Gotcha! Not sure what was going on last week, but can now reproduce this. Looks like it's a known issue and documented here https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#flake8

adamdougal (approved):
LGTM! Thanks for this!

PR #1395, opened 2024-10-09
fix: check test case failing issue

Rohini-Microsoft:
## Purpose

* ...

- [ ] Yes
- [ ] No

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...


Rohini-Microsoft:
this is created for testing purpose so closed it

PR #1460, opened 2024-11-04
fix: ci workflow fix for Bicep deployment

Roopan-Microsoft:
## Purpose

* Fix provided for CI workflow where resource group is creating automatically from the GitHub variables.

- [ ] Yes
- [x] No

## What to Check
CI workflow - Automated


Avijit-Microsoft (approved):

ross-p-smith:
@Roopan-Microsoft - I dont think this will fix the underlying problem. We have now added an extra step that developers have to do - create a resource group. Why has the creation of the RG been removed from the arm/bicep?

PR #1538, opened 2024-11-27
fix: Auth type set default to rbac and removed resource group from bicep

Roopan-Microsoft:
## Purpose

* Auth type set default to RBAC
* Reverted Resource group scope from make file and yml files

- [ ] Yes
- [x] No

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/infra/main.json
+++ b/infra/main.json
@@ -4,8 +4,8 @@
   "metadata": {
     "_generator": {
       "name": "bicep",
-      "version": "0.29.47.4906",
-      "templateHash": "5982658886499159314"
+      "version": "0.31.92.45157",
+      "templateHash": "18364921757248167990"
     }
   },
   "parameters": {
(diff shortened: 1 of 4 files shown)


infra/main.json
```diff
@@ -4,8 +4,8 @@
   "metadata": {
     "_generator": {
       "name": "bicep",
-      "version": "0.31.34.60546",
-      "templateHash": "9243131736696562084"
+      "version": "0.29.47.4906",
```
ross-p-smith, line 7: Why have we downgraded the version of bicep - this should have only ever gone up. This makes me feel that bicep has not been updated on your local machine.

Roopan-Microsoft: Thanks for identifying @ross-p-smith . Its fixed now.

Prajwal-Microsoft (approved):

PR #1544, opened 2024-11-28
fix: Bug-10777| LLM Model Responds in Spanish

Roopan-Microsoft:
## Purpose

* Fixed: LLM Model Responds in Spanish when we ask questions in english
* Now it will keep default language as English and reply in the language what user speaks.

- [ ] Yes
- [x] No

## How to Test
*  Get the code
* Deploy
* Ask questions in different languages

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...


marktayl1 (approved):
I will approve changes to prompting. All prompting seems only related to language choice and does not appear to impact the scenario outside of this.

PR #1545, opened 2024-11-28
fix: Bug-9496 | Response handled if we ask suggested questions for the tool

Roopan-Microsoft:
## Purpose

* Response handled if we ask suggested questions for the tool
* It will respond like for questions like "Give me some suggested questions for this tool?""

- [ ] Yes
- [x] No

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* Check by asking questions like "Can you suggest some question I should ask to this tool?"


marktayl1 (approved):
I will approve changes to prompting. My understanding is that this prompting is designed to ground the model and only applies to grounding the model in its data.

PR #1561, opened 2024-12-18
Fix for bug 11703 - conversational flow ('BYOD' instead of defaulting to 'Custom')

Priyanka-Microsoft:
…ected conversational flow ('BYOD' instead of defaulting to 'Custom')

## Purpose

* ...

- [ ] Yes
- [ ] No

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...


Priyanka-Microsoft:
Creating another PR for this

PR #1627, opened 2025-01-14
feat: Support rbac auth type in conversation_logger

leonavevor:
## Purpose

* To support rbac when conversation history is being saved 

- [ ] Yes
- [x] No

## How to Test
*  Run it in devcontainer to test

```
git clone [[repo-address]](https://github.com/leonavevor/chat-with-your-data-solution-accelerator)
cd chat-with-your-data-solution-accelerator
git checkout main
```

* Test the code

```
code/tests/utilities/helpers/test_azure_search_helper.py => test_get_conversation_logger_rbac()
```

## What to Check
* should be able to have your conversations saved even if you are not using key based auth (i.e using RBAC).


code/backend/batch/utilities/helpers/azure_search_helper.py
```diff
@@ -265,15 +265,24 @@ def get_conversation_logger(self):
...
-                if self.env_helper.is_auth_type_keys()
-                else None
-            ),
-            index_name=self.env_helper.AZURE_SEARCH_CONVERSATIONS_LOG_INDEX,
-            embedding_function=self.llm_helper.get_embedding_model().embed_query,
-            fields=fields,
-            user_agent="langchain chatwithyourdata-sa",
-        )
+        if self.env_helper.AZURE_AUTH_TYPE == "rbac":
+            return SearchIndexClient(
```
Prasanjeet-Microsoft, line 269: In the RBAC condition, the current implementation returns a SearchClient object from the Azure SDK, which does not support the .add_texts() method used for logging conversations. This results in a runtime AttributeError.
To resolve this, we should ensure that the RBAC path also returns a LangChain-compatible AzureSearch instance to maintain consistent behavior across both auth types.

Please find the below screenshot of the error for your reference:
image

Prasanjeet-Microsoft: > In the RBAC condition, the current implementation returns a SearchClient object from the Azure SDK, which does not support the .add_texts() method used for logging conversations. This results in a runtime AttributeError. To resolve this, we should ensure that the RBAC path also returns a LangChain-compatible AzureSearch instance to maintain consistent behavior across both auth types.
> 
> Please find the below screenshot of the error for your reference: image

Hello @leonavevor, just following up on the comment I left earlier regarding the issue I noticed in this PR. Let me know if you need any clarification or if there's anything I can help with to move this forward. Thanks!

leonavevor: @Prasanjeet-Microsoft Let i will take a look at that and make the necessary change to fix it ASAP. I will push a patch soon

leonavevor:
@Prasanjeet-Microsoft and @Vinay-Microsoft I have committed a better fix for this issue and rebased my recent changes on it, so it is in sync with the upstream main.
This fix is base on AzureSearch library.
This is the new change: https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/pull/1627/commits/92e7f9847e7b00ae277e1191c2d1acb45d3c1dbb

Prasanjeet-Microsoft:
> @Prasanjeet-Microsoft and @Vinay-Microsoft I have committed a better fix for this issue and rebased my recent changes on it, so it is in sync with the upstream main. This fix is base on AzureSearch library. This is the new change: [92e7f98](https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator/commit/92e7f9847e7b00ae277e1191c2d1acb45d3c1dbb)

Hello @leonavevor, Thanks for the updated fix and for incorporating the changes using the AzureSearch library. We will verify the changes and if everything works as expected, we will proceed with merging the PR.

Roopan-Microsoft (approved):

PR #1711, opened 2025-03-14
refactor: Config attribute classes to use BaseConfig with get methods

vediyd-calab:
## Purpose

1. **Created `BaseConfig` class:**
    - Added `__getitem__` method for dictionary-like access.
    - Added `get` method for safe attribute access with a default value.

2. **Updated classes to inherit from `BaseConfig`:**
    - `Config`
    - `Prompts`
    - `Example`
    - `Messages`
    - `Logging`
    - `IntegratedVectorizationConfig`

3. **Added type hints using `typing` module:**

- [ ] Yes
- [x] No

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

NOTE: Haven't tested the code since the changes seem minor. 

## What to Check
* `code/backend/batch/utilities/helpers/config/config_helper.py`

## Other Information

The changes aim to:
- Simplify the code by centralizing common functionality in a base class.
- Improve code readability and maintainability.
- Leverage type hints for better static analysis and code understanding.
- As the solution becomes more complex on the admin app, this config and its sub-attributes start having variables which are a little too long. Having a `get` function would help handle calling specific attributes within, e.g.:
  ```python
  ASP = "answering_system_prompt"
  config.prompts.get(ASP, "")
  ```
  Since the `get` function doesn't exist, the lines get too long.

### DIFF
--- a/code/backend/batch/utilities/helpers/config/config_helper.py
+++ b/code/backend/batch/utilities/helpers/config/config_helper.py
@@ -3,6 +3,7 @@ import json
 import logging
 import functools
 from string import Template
+from typing import List, Dict, Optional
 
 from ..azure_blob_storage_client import AzureBlobStorageClient
 from ...document_chunking.chunking_strategy import ChunkingStrategy, ChunkingSettings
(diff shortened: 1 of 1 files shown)


vediyd-calab:
For now using `getattr` works fine without requiring this change here. Closing.

